PR #92, opened 2026-07-03
fix(adlc-init): track .adlc/specs/, exclude .adlc/ from formatters, warn on ticket-write quality checks

### LINKED ISSUE #42: adlc-init: exclude .adlc/ from repo formatters/linters (frozen tickets.json trips Biome/Prettier)
## Context

## Observed
`.adlc/tickets.json` is machine-written (`JSON.stringify(…, null, 2)`, multi-line arrays). The repo formatter (Biome via ultracite) wanted `"rails": ["lib/content.ts"]` inline and failed `pnpm check` in CI — turning `main` red the moment the ticket was committed.

The trap: once a ticket declares `rails`, `tickets.json` becomes the frozen trust root, so it **cannot be reformatted on a ticket branch** without tripping `rails-guard`. The only clean fix was excluding `.adlc/` from the formatter.

## Proposed fix
`adlc-init` should add `.adlc/` to detected formatter/linter ignore configs during bootstrap (biome overrides, `.prettierignore`, eslint ignores) — mirroring how it already manages `.gitignore`. Alternatively emit formatter-stable JSON. At minimum, document the conflict.

### LINKED ISSUE #43: /adlc-ticket: run the repo quality check after writing tickets.json and warn on failure
## Context

## Observed
Committing `tickets.json` silently turned `main` CI red (the formatter conflict, see sibling issue). Nothing in the authoring flow caught it; it was discovered only on the next PR.

## Proposed fix
After the atomic write in `/adlc-ticket`, detect the repo's quality script (`check`/`lint`) and run it against `.adlc/tickets.json`, warning if it fails — so a ticket write can't silently red the build.

### LINKED ISSUE #46: adlc-init: track P1 specs (.adlc/specs/) instead of leaving them git-ignored
## Context

## Observed
The P1 spec lives at `.adlc/specs/T1.md` — arguably the most important contract in the lifecycle — but it is **untracked**, because `adlc-init`'s `.gitignore` only negates `tickets.json` (`.adlc/*` + `!.adlc/tickets.json`). The stress-tested spec that drove the build isn't versioned by default.

## Proposed fix
Decide that specs are a tracked contract and have `adlc-init` also negate `.adlc/specs/` (`!.adlc/specs/`), or relocate specs outside the runtime-ignored tree. Document the choice either way.

voodootikigod:
Fixes #46
Fixes #42
Fixes #43

## Summary

`adlc-init` (cursor + opencode plugins) had three related hygiene gaps surfaced by running a full P0→P6 cycle on an external repo:

- **#46** — `.adlc/specs/` (the P1 spec, arguably the most important contract in the lifecycle) was left git-ignored because the scaffolded `.gitignore` stanza only negated `.adlc/tickets.json`. `ensureGitignore()` now also negates `!.adlc/specs/`.
- **#42** — a machine-written `.adlc/tickets.json` (multi-line `JSON.stringify`) conflicts with repo formatters/linters (Biome, Prettier, ESLint) — and once a ticket declares `rails`, `tickets.json` becomes a frozen trust root that can't be reformatted on a ticket branch without tripping rails-guard. Added `ensureFormatterIgnores()`, which detects and patches `biome.json`, `.prettierignore`, `.eslintrc.json`/`ignorePatterns`, `.eslintignore`, and reports (but does not silently mutate) flat `eslint.config.js`.
- **#43** — `/adlc-ticket` now runs the repo's detected quality/check script against `.adlc/tickets.json` after the atomic write and warns (non-blocking) on failure, instead of silently letting a bad write turn CI red on the next PR.

Both `adlc-cursor` and `adlc-opencode` carry byte-for-byte duplicated `ensureGitignore`/`ensureFormatterIgnores` implementations (matching the existing duplication pattern in these packages), so every fix below was applied and tested identically in both.

### Review history (5 rounds, iterative hardening of `ensureGitignore`)

Adversarial review surfaced a sequence of git `.gitignore` last-match-wins edge cases in `ensureGitignore`, each fixed in its own commit:

1. Duplicating a standalone negation line when the `.adlc/*` anchor was absent.
2. Negation line ending up *before* a newly-appended anchor when the anchor was absent (silently re-ignoring the negated file despite reporting `changed: true`).
3. A negation line sitting before an *existing* anchor (hand-edit/merge) was never relocated — an early-return masked this because "all lines present, just misordered" looked like "already correct."
4. A duplicate `.adlc/*` anchor (second occurrence from a merge/second tool run) silently re-ignored the negation while reporting `changed: false`. Also: `ensureFormatterIgnores` only reported one of `.eslintrc.json` / `.eslintignore` when both existed, dropping the other mutation from the CLI summary.
5. The CLI printed a misleading "updated (added )" summary when a repair changed nothing in `added` (pure anchor/order repair, no new stanza lines).

## NEEDS ATTENTION — review did not reach a clean verdict

This branch went through **5 rounds of adversarial review and did not reach a clean (0-finding) verdict**:

| Round | Findings |
|-------|----------|
| 1 | 2 |
| 2 | 1 |
| 3 | 2 |
| 4 | 4 |
| 5 | 1 |

Every reported finding through round 5 has a corresponding fix commit (see history above and `git log` on this branch), and the round-5 fix is included in the tip commit (`cb0e72a`). **However, the round-5 fix itself was not re-run through a subsequent clean review round** (the process stopped at the round cap), so it has not been adversarially re-verified.

Please have a human reviewer specifically check before merging:

1. **No further `.gitignore` last-match-wins edge cases remain.** Five rounds in a row found new variants of the same class of bug (anchor absent / misordered / duplicated, negation before/after, partial vs. full stanza). Consider whether the fix is now exhaustive or whether a property-style/fuzz test over anchor+negation permutations would be more convincing than the current example-based regression tests.
2. **The duplicated `ensureGitignore`/`ensureFormatterIgnores` logic across `adlc-cursor` and `adlc-opencode`.** Every fix in this PR had to be applied twice (once per package) to keep them in sync. This PR intentionally did not extract a shared module (per the round-4 commit message, treating that as a separately-scoped consolidation), so a future change to one copy without the other will silently reintroduce drift. Worth a follow-up ticket.
3. **`ensureFormatterIgnores` flat `eslint.config.js` handling** — it only *detects and reports*, it does not mutate. Confirm that's the intended scope for this PR (vs. actually patching flat config).
4. **#43's quality-check-after-write is advisory/warn-only**, not blocking — confirm that matches the issue's intent (issue proposes "warning if it fails," which this implements, but flag for reviewer sign-off given the CI-red history that motivated the issue).

## Test plan

- [x] `node --test plugins/adlc-cursor/test/*.test.mjs` — 54/54 pass
- [x] `node --test plugins/adlc-opencode/test/*.test.mjs` — 71/71 pass
- [x] Full repo suite (`npm test`, covers all packages/plugins/apps) — green, exit 0
- [x] Regression tests added for every round-1–5 finding (duplicate negation, misordered negation, duplicate anchor, dual eslint source reporting, CLI "added" summary wording)
- [ ] Human re-review of the round-5 fix and the open concerns above before merge

### DIFF
--- a/docs/specs/adlc-init-hygiene.md
+++ b/docs/specs/adlc-init-hygiene.md
@@ -1,0 +1,81 @@
+# Spec — adlc-init hygiene sweep (#46, #42, #43)
+
+**Phase:** P1 record for the `adlc-init-hygiene` cluster. Three related hygiene fixes to
+the per-harness `/adlc-init` bootstrap and the `/adlc-ticket` authoring flow, surfaced by
+running a full P0→P6 ADLC cycle on an external repo.
+
+## Issue #46 — track `.adlc/specs/` by default
+
+**Problem:** every `/adlc-init`'s documented `.gitignore` stanza was `.adlc/*` +
+`!.adlc/tickets.json` only, so the P1 spec at `.adlc/specs/T<n>.md` — arguably the most
+important contract in the lifecycle — was left untracked by default.
+
+**Acceptance criteria:**
+- Every harness's `/adlc-init` (`adlc-claude-code`, `adlc-opencode`, `adlc-antigravity`,
+  `adlc-cursor`) documents a three-line stanza: `.adlc/*`, `!.adlc/tickets.json`,
+  `!.adlc/specs/`.
+- Where a deterministic scaffolder exists (`adlc-opencode`, `adlc-cursor`), `ensureGitignore()`
+  in `lib/scaffold.mjs` writes/repairs this stanza: creates it fresh if `.gitignore` is
+  absent, and adds only the missing `!.adlc/specs/` line if the older two-line form is
+  already present, leaving every other line untouched.
+
+**Verify:**
+```sh
+node --test plugins/adlc-opencode/test/scaffold.test.mjs
+node --test plugins/adlc-cursor/test/scaffold.test.mjs
+```
+
+## Issue #42 — exclude `.adlc/` from repo formatters/linters
+
+**Problem:** `.adlc/tickets.json` is machine-written (`JSON.stringify(…, null, 2)`). Once a
+ticket declares `rails`, the file becomes a frozen trust root that cannot be reformatted on
+a ticket branch without tripping `rails-guard` — so a repo formatter reformatting it turns
+CI red on the very next PR.
+
+**Acceptance criteria:**
+- Every harness's `/adlc-init` documents detecting and updating, only for configs already
+  present in the target repo: Biome (`biome.json` → merged `overrides` entry disabling
+  formatter/linter for `.adlc/**`), Prettier (`.prettierignore` → appended `.adlc/` line),
+  ESLint (`ignorePatterns` in a JSON `.eslintrc*`, or `.eslintignore`; flat
+  `eslint.config.js`/`.mjs`/`.cjs` is detected but not auto-edited).
+- The manual fallback is documented for tools not auto-detected (flat ESLint config, or any
+  other formatter/linter).
+- `lib/scaffold.mjs`'s `ensureFormatterIgnores()` (opencode, cursor) implements this
+  deterministically and is wired into `scaffold()`.
+
+**Verify:**
+```sh
+node --test plugins/adlc-opencode/test/scaffold.test.mjs
+node --test plugins/adlc-cursor/test/scaffold.test.mjs
+```
+
+## Issue #43 — warn if the quality check flags `tickets.json` after a write
+
+**Problem:** nothing in the `/adlc-ticket` authoring flow caught the formatter conflict
+above; it was discovered only on the next PR, after `.adlc/tickets.json` had already been
+committed.
+
+**Acceptance criteria:**
+- Both harnesses that ship `/adlc-ticket` (`adlc-claude-code`, `adlc-opencode`) document a
+  step, run immediately after the atomic write and lock release, that: looks for
+  `scripts.check` then `scripts.lint` in `package.json`; detects the package manager from
+  the lockfile; runs that check scoped to `.adlc/tickets.json` where the underlying tool
+  supports a path argument, else the full script; and on failure **warns** (never fails or
+  reverts the write) with a pointer back to `/adlc-init`'s formatter-ignore step.
+- No script found → skip silently; check passes → continue silently.
+
+**Verify:** manual read-through of `plugins/adlc-claude-code/commands/adlc-ticket.md` §3 and
+`plugins/adlc-opencode/command/adlc-ticket.md` §4 (this is an agentic instruction, not a
+deterministic script — there is no unit test to run).
+
+## Full regression check
+
+```sh
+node --test plugins/adlc-opencode/test/*.test.mjs
+node --test plugins/adlc-cursor/test/*.test.mjs
+node --test plugins/adlc-antigravity/test/*.test.mjs
+node --test plugins/adlc-claude-code/hooks/test/*.test.mjs
+node scripts/cursor-install-smoke.mjs .
+node scripts/opencode-install-smoke.mjs .
+node scripts/antigravity-install-smoke.mjs .
+```
--- a/plugins/adlc-antigravity/commands/adlc-init.md
+++ b/plugins/adlc-antigravity/commands/adlc-init.md
@@ -16,15 +16,36 @@ Bootstrap the ADLC runtime for use with `agy`.
    ```sh
    adlc init || npx @adlc/cli init
    ```
-3. **Add the .gitignore stanza** so only the ticket file is tracked:
+3. **Add the .gitignore stanza** so the ticket file and P1 specs are tracked,
+   everything else under `.adlc/` is not:
    ```
    .adlc/*
    !.adlc/tickets.json
+   !.adlc/specs/
    ```
-4. **Wire the CI gate** (the real guarantee): copy `docs/ci/rails-guard.yml` into your
+   If `.gitignore` already has the older two-line form (no `!.adlc/specs/`),
+   add just the missing `!.adlc/specs/` line — don't touch anything else. The
+   `specs/` directory holds the P1 spec for every ticket, arguably the most
+   important contract in the lifecycle, so it must be tracked by default.
+4. **Exclude `.adlc/` from this repo's formatters/linters.** `.adlc/tickets.json`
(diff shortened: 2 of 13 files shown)


voodootikigod:
## Fresh adversarial pass (round 6) — rebase + re-verification

**Rebase**: rebased cleanly onto `origin/main` (a265575). The one conflict, in
`plugins/adlc-antigravity/commands/adlc-init.md` (this branch's formatter-ignore
step vs. PR #78's fail-open CI-gate callout), was resolved additively — both are
now present: step 4 (exclude `.adlc/` from formatters) and step 5 (wire the CI
gate, with the fail-open `scripts/rails-guard-ci.mjs` callout from #78 intact),
renumbered through step 6. Force-pushed (`cb0e72a` → `47ea91a`).

### 1. Round 5's CLI wording fix — confirmed correct

Ran the actual CLI against a live `.gitignore` with only a duplicate `.adlc/*`
anchor (a pure repair, no new stanza lines):

```
.gitignore             — updated (repaired duplicate/misplaced entries)
```

Confirmed all three branches render correctly end-to-end in both packages
(cursor + opencode): no-op ("already tracks..."), repair-only (new wording,
no bogus "added " with nothing after it), and genuine additions
("added .adlc/*, !.adlc/tickets.json, !.adlc/specs/"). No issue found.

### 2. One more `.gitignore` permutation — none found; fuzz-tested instead

Rather than hand-construct a 6th variant, I wrote a property/fuzz harness (per
the PR body's own suggestion) and ran it against the real `ensureGitignore()`
in both packages: **150,000+ generated trials** combining 0–5 copies of the
`.adlc/*` anchor, 0–4 copies of each negation line, random noise lines
(including near-miss patterns like `.adlc.bak/*` and `!.adlc/other.json`),
random shuffled ordering, and with/without a trailing newline — asserting on
every trial that (a) `tickets.json`/`specs/` end up tracked (simulated
last-match-wins), (b) at most one anchor survives, (c) a second call is a
true no-op (idempotent), and (d) an unchanged file is never rewritten.
**Zero failures** in both `adlc-cursor` and `adlc-opencode`.

**Verdict: the 5 rounds already covered the real edge-case space** for the
literal 3-line stanza — the fix correctly reduces to "one anchor + both
negations somewhere after it" regardless of how the input is scrambled, and
that's the entire state space for this specific string-based approach.

One related but genuinely *different* class of risk exists outside this
scope: `ensureGitignore()` only recognizes its own literal `.adlc/*` /
`!.adlc/tickets.json` / `!.adlc/specs/` strings. A **pre-existing, differently
worded** ignore entry elsewhere in the file — e.g. a hand-written bare
`.adlc/` (directory-level, no `*`) or `.adlc/**` — would not be detected or
reconciled, and directory-level ignores have a separate git gotcha (git won't
descend into an ignored directory to re-include a child file at all, no
matter where the negation sits). This isn't a permutation of the anchor the
function manages, and fixing it well requires pattern-equivalence reasoning,
not another ordering fix — recommend a separate follow-up ticket rather than
blocking this PR on it.

### 3. Flat `eslint.config.js` detect-only — confirmed intentional, already documented

Confirmed this is a deliberate scope decision, not an oversight, in four
places: the code comment above `ensureEslintRcIgnore` ("executable JS — safe
text mutation isn't reliable, so only report detection"), the
`ensureFormatterIgnores` JSDoc, both `adlc-init.md` command docs
(cursor: "only *detected*, not auto-edited"; opencode: same, "it's executable
JS"), and a regression test that asserts the file is byte-for-byte untouched
after the call. No change needed.

### Test results

- `node --test plugins/adlc-cursor/test/*.test.mjs` — 54/54 pass
- `node --test plugins/adlc-opencode/test/*.test.mjs` — 71/71 pass
- No logic changes were made (fuzz testing found nothing to fix), so no new
  regression tests were added beyond the existing round 1–5 coverage.

### Recommendation on `[NEEDS ATTENTION]`

Of the 4 open concerns listed in the PR body:

1. Gitignore edge-case exhaustiveness — **resolved** (fuzz-verified above).
2. `ensureGitignore`/`ensureFormatterIgnores` duplication across packages —
   **already accepted as-is** by the requester; follow-up consolidation
   ticket to be filed separately (out of scope for this pass, per explicit
   instruction).
3. Flat `eslint.config.js` detect-only scope — **resolved**, confirmed
   intentional and documented.
4. #43's quality-check-after-write being advisory/warn-only — **not
   evaluated in this pass** (outside this review's assigned scope); still
   needs a human sign-off call against the issue's original intent.

Given 1 and 3 are now closed and 2 has already been explicitly accepted, the
only remaining open item is 4, which is a product-intent question rather than
a correctness bug. **The `[NEEDS ATTENTION]` tag can reasonably be downgraded
to "final human sign-off on #43's warn-vs-fail intent"** rather than a broad
"review didn't reach clean verdict" — but I'd leave that judgment call, and
whether to rename the PR title, to a human reviewer since #43 wasn't part of
my assigned scope here.

voodootikigod:
**Follow-up**: `main` advanced again (PR #89, the `/adlc:adlc-<name>` namespacing
fix) while this review pass was in flight, so I rebased a second time onto the
new tip (`fa3b8bd`). One additional conflict surfaced in
`plugins/adlc-claude-code/commands/adlc-init.md` (summarize-step wording,
resolved additively — kept both the formatter/linter summary text from this PR
and the `/adlc:adlc-ticket` namespacing from #89).

More importantly, that rebase exposed a real defect: this PR's new step 3 in
`plugins/adlc-claude-code/commands/adlc-ticket.md` (the #43 warn-on-write-quality
guidance) referenced the bare, non-namespaced `` `/adlc-init` `` twice — exactly
the defect class #89's new `claude-code-plugin-smoke.mjs` regression guard
exists to catch. Fixed both to `` `/adlc:adlc-init` ``. Re-ran the smoke test
and its own regression suite after fixing:

- `node scripts/claude-code-plugin-smoke.mjs` — passes (no `fail()`, exit 0)
- `node --test scripts/test/claude-code-plugin-smoke.test.mjs` — 5/5 pass
- `node --test plugins/adlc-cursor/test/*.test.mjs` — 54/54 pass (re-verified)
- `node --test plugins/adlc-opencode/test/*.test.mjs` — 71/71 pass (re-verified)

Force-pushed again (`47ea91a` → `49335d4`). `gh pr view 92` now reports
`mergeable: MERGEABLE` (the `BLOCKED` status is pending/required CI checks, not
a conflict).

PR #93, opened 2026-07-03
feat(build-gate): machine-checkable fitness-to-build gate

### LINKED ISSUE #48: ADLC: machine-checkable "fitness-to-build" gate — block high-blast-radius builds in a degraded context
## Problem

ADLC gates the *quality* of a change (coldstart, prosecute/P5) but nothing **deterministically prevents starting** a high-blast-radius ticket's build (P4) when the executing context is degraded ("context rot" — a model failure mode ADLC explicitly names). On 2026-06-27, building `@adlc/ticket-sync`, the right call (pause before T9's create/id-reassignment logic) happened **only because the operator asked "should we pause?"** — i.e. it relied on human/model judgment, the exact anti-pattern ADLC exists to remove ("if it traces to a model failure mode, make it a machine-checkable gate, not a thing to remember").

T9's deferred logic is the motivating example: idempotent `gh issue create` + id reassignment (`T<n>`→`gh:owner/repo#n`) with store-wide edge rewrite and manifest-evidence migration. Wrong, it **duplicates real issues or orphans dependency edges** — high, partly-irreversible, outward blast radius. Building that in a deep/degraded context is where confident-hallucination bugs are most likely, and per-change P5 then pays for it in many BLOCKED prosecution rounds.

## Proposal — a "fitness-to-build" gate (P3 → P4 entry gate)

A deterministic gate (`adlc build-gate <ticket>`, plus a Claude Code PreToolUse/SessionStart hook) that refuses to start a **high-risk** build in a degraded context unless an audited override is recorded.

1. **Risk tier per ticket** — declared `risk: high|normal` or derived from signals already in the ticket: writes back to / creates / deletes in an external system; mutates identity or `.adlc/manifest.jsonl`; touches the trust root; `category` in {contract, architecture}. High = irreversible/outward/identity-mutating.
2. **Context-fitness signal** — a machine-checkable proxy for "this session is deep." `flail-detector` already scans the session transcript with a byte cap; reuse that — transcript size / tool-call count crossing a threshold = degraded. The CC plugin's hooks already receive `transcript_path`, so this is observable in-session (the CLIs can take `--session-bytes`/`--depth`).
3. **Enforcement** — for a high-risk ticket, **deny starting P4** when the fitness signal is past threshold, unless an **audited override** is recorded to the gate-manifest (mirrors `ADLC_RAILS_BYPASS`: an un-auditable override is refused). The denial message recommends resuming in a fresh session / isolated subagent.
4. **Why a cold resume is safe** — `coldstart` already certifies tickets are self-contained and executable with no conversation context; rails freeze the upstream contract; upstream tickets are P5-prosecuted. So "pause and resume cold" loses nothing — this gate makes that the default safe path for high-risk work.

### Tie-ins
- `model-router` already tiers tickets; extend it to emit a **"fresh-context required"** flag for high-risk tickets.
- Pairs with the SessionStart "ADLC precedence" assertion (see #35 section C) — same theme: make ADLC self-assert via the harness, not via operator memory.

## Acceptance (sketch)
- [ ] Risk tier derivable + overridable on a ticket
- [ ] A context-fitness signal exposed to the gate (transcript-size proxy via the CC hook; `--depth` input for the CLI)
- [ ] `build-gate` denies a high-risk build past threshold; audited override via gate-manifest; exit `2` on deny
- [ ] CC PreToolUse hook enforces it in-session for declared high-risk tickets
- [ ] Docs + tests (offline; injected signal)

## Context
Surfaced during the `@adlc/ticket-sync` dogfooding build (PR #33). Companion to the dogfooding backlog in #35 (this is the concrete, escalated form of the "make ADLC self-assert" theme). Highest-leverage because it defends the lifecycle's *entry* into its riskiest work.

## Implementation paths — handle in EITHER (or both)

The risk-tier model + gate semantics are shared; the *context-fitness signal* and *enforcement point* differ by surface. Either path delivers value independently; both give full coverage.

### Path A — Toolkit (`@adlc` CLI) — harness-agnostic
- `adlc build-gate <ticket> --depth <n>` (or `--session-bytes`): deterministic gate; risk derivation lives in core/`model-router`; deny = exit `2`; audited override via `gate-manifest`.
- Pro: works for **every** harness (Codex, Pi, CI), not just Claude Code; testable offline (inject the depth signal).
- Con: a CLI process can't observe context state itself — the caller (hook/wrapper/CI) must pass the `--depth`/`--session-bytes` signal in.

### Path B — Claude Code plugin (hook) — self-enforcing in-session
- A PreToolUse/SessionStart hook (the plugin already receives `transcript_path`, like `flail-detector`): compute the fitness signal from transcript size/tool-count itself, look up the active ticket's risk tier, and **deny Edit/Write** for a high-risk ticket past threshold (audited override env var, mirroring `ADLC_RAILS_BYPASS`).
- Pro: zero operator action — the harness self-asserts; reuses existing rail-hook plumbing.
- Con: Claude Code only.

### Recommended composition
Toolkit owns risk-tier derivation + `build-gate` + audited override (the reusable core); the CC hook supplies the in-session context signal and the enforcement point by shelling to `adlc build-gate --depth <transcript-derived>`. Ship Path B first (self-enforcing, highest leverage in CC), with the risk-tier/gate logic factored into the toolkit so other harnesses can adopt Path A.

voodootikigod:
Fixes #48

## Summary

Adds `@adlc/build-gate`: a machine-checkable P3→P4 fitness-to-build gate that denies **starting** a high-risk ticket's build when the executing session's context is degraded (context rot), unless an audited override is recorded. Previously nothing prevented this deterministically — it relied on an operator noticing and asking to pause.

**Toolkit (Path A, reusable by any harness)** — `packages/build-gate`:
- `lib/risk.mjs` — risk-tier derivation: declared `risk: 'high'`, or derived from category in `{contract, architecture}`, external-system effects, identity mutation, or scope/rails touching the trust root / gate-manifest ledger. A declared `risk: 'normal'` can never downgrade a derived-high signal.
- `lib/depth-signal.mjs` — the context-fitness proxy (tool-call depth + transcript bytes), mirroring `@adlc/flail-detector`'s existing transcript-size/tool-call-count scanning approach.
- `lib/decide.mjs` — the pure allow/deny decision.
- `lib/override.mjs` — durable override recording, routed through `@adlc/gate-manifest`'s real chain builder so bypass entries are hash-chain-linked and signed like every other ledger entry.
- `lib/active-ticket.mjs` — the shared `ADLC_TICKET`/`.adlc/current-ticket.json` active-ticket convention already used by every other harness integration.
- `bin/build-gate.mjs` — `adlc build-gate <ticket-id> [--depth n] [--session-bytes n] [--transcript path] [...]`, registered in `packages/cli/lib/registry.mjs`.

**Claude Code hook (Path B, self-enforcing, the primary deliverable)** — a new `buildgate` PreToolUse mode in `plugins/adlc-claude-code/hooks/adlc-hook.mjs`, wired alongside `rails` in `hooks.json`. It resolves the active ticket and computes risk tier + context-fitness entirely locally (verbatim KEEP-IN-SYNC ports of the toolkit logic), denying the edit when the ticket is high-risk and the session is degraded — unless `ADLC_BUILD_GATE_BYPASS=1` is set **and** the override is durably recorded via `adlc gate-manifest record build-gate-bypass` (mirrors `ADLC_RAILS_BYPASS`: an unauditable override is refused, never silent). The core decision does not depend on the `adlc` CLI being on `PATH`, so installing the plugin cannot brick a session that hasn't opted into the active-ticket convention.

### Review history (5 rounds)

| Round | Findings | Fix |
|-------|----------|-----|
| 1 | 2 | `b5dffda` — fail closed when a high-risk ticket's `transcript_path` exists but reading it fails (permission error / TOCTOU race) instead of silently treating it as zero bytes; freeze `.adlc/current-ticket.json` as a rails trust root (mirroring `.adlc/tickets.json`) so a structured edit can't silently repoint the active-ticket pointer for a ticket that declares rails. |
| 2 | 1 | `a5f0489` — `recordOverride()` was appending a raw manifest entry directly instead of going through `@adlc/gate-manifest`'s real chain builder, which broke the hash chain (no `seq`/`prev`, wrong shape) and made `gate-manifest verify` falsely report tampering after a legitimate bypass. Now routes through `gate-manifest`'s `record()`. |
| 3 | 1 | `0355d75` — `deriveRiskSignals()` spread a ticket's `scope`/`rails` fields with no `Array.isArray` guard; a malformed field (string glob, number, object) threw an uncaught `TypeError` and crashed the CLI. Now fails closed: a malformed field drives the tier to `high` (`malformed-scope`/`malformed-rails`) instead of crashing or being silently dropped. |
| 4 | 0 | — |
| 5 | 1 | `8433d86` — corrected a README claim in `docs/integrations/claude-code.md` stating the buildgate hook "shells to `adlc build-gate <id>`" for its decision; it doesn't — the risk-tier/context-fitness/decide logic is computed entirely locally via the KEEP-IN-SYNC ports. `adlc` is only invoked to record an audited bypass. |

## NEEDS ATTENTION — review did not reach a clean verdict

This branch went through **5 rounds of adversarial review and did not conclude with a clean (0-finding) round** — round 5 found 1 finding, which was fixed in the tip commit (`8433d86`), but that fix was **not re-run through a subsequent clean review round** (the process stopped at the round cap). Please have a human reviewer specifically check before merging:

1. **The round-5 fix itself is unverified by re-review.** `8433d86` is a docs-only correction (no code change), so the blast radius should be small, but confirm no other doc/behavior claims in `docs/integrations/claude-code.md` or `docs/specs/build-gate-fitness.md` still overstate what the hook actually checks.
2. **Bash bypasses the gate entirely (self-documented, not fixed).** The PreToolUse hook only matches structured-edit tools (`Edit`/`Write`/`MultiEdit`/`NotebookEdit`) — never `Bash` — same documented limitation as `rails-guard`. `rm .adlc/current-ticket.json` or any out-of-band overwrite via Bash clears the active-ticket pointer with zero risk evaluation and zero manifest entry. See `docs/specs/build-gate-fitness.md`'s "Known limitations" section.
3. **A high-risk ticket with no declared `rails` has no trust-root protection at all.** The `.adlc/current-ticket.json` freeze only activates when the ticket set declares at least one rail; a ticket that's high-risk purely via `category: 'contract'`/`architecture` (no rails) gets no trust-root protection, and — unlike rails' Bash gap — there is **no CI diff backstop** possible here, since `.adlc/current-ticket.json` is gitignored local session state, not tracked. Confirm this residual gap is acceptable for merge, or whether it needs a follow-up ticket.
4. **Unreadable `transcript_path` fails closed by design** — confirm the team is comfortable that a permission error / TOCTOU race on the transcript file denies the edit outright (rather than degrading gracefully), since this is a hard "unverifiable session must not be allowed through" stance with no fallback.

Treat build-gate as a strong in-session backstop, not a substitute for human review at P5/P6 on high-risk tickets — this is called out explicitly in the spec doc.

## Test plan

- [x] `node --test packages/build-gate/test/*.test.mjs` — 67/67 pass
- [x] `node --test packages/cli/test/*.test.mjs` — 14/14 pass
- [x] `node --test plugins/adlc-claude-code/hooks/test/*.test.mjs` — 69/69 pass
- [x] Full repo suite (`npm test`, covers all packages/plugins/apps) — green, exit 0, zero failures across all 32 test files
- [x] Regression tests added for every round 1–5 finding (TOCTOU transcript read failure, current-ticket.json trust-root freeze, gate-manifest chain-linked override recording + `verify()` after append, non-array scope/rails fail-closed)
- [ ] Human re-review of the round-5 fix and the open concerns above before merge

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,7 +47,7 @@ hooks, CI, and lifecycle docs should use the dispatcher.
 | Phase | Packages |
 | --- | --- |
 | **Spec & ticket shaping** | [`parallax`](./packages/parallax) · [`spec-lint`](./packages/spec-lint) · [`premortem`](./packages/premortem) · [`coldstart`](./packages/coldstart) · [`ticket-prune`](./packages/ticket-prune) |
-| **Execution supervision & rails** | [`preflight`](./packages/preflight) · [`model-router`](./packages/model-router) · [`merge-forecast`](./packages/merge-forecast) · [`rails-guard`](./packages/rails-guard) · [`flail-detector`](./packages/flail-detector) · [`consensus-fix`](./packages/consensus-fix) · [`runner`](./packages/runner) |
+| **Execution supervision & rails** | [`preflight`](./packages/preflight) · [`model-router`](./packages/model-router) · [`merge-forecast`](./packages/merge-forecast) · [`rails-guard`](./packages/rails-guard) · [`build-gate`](./packages/build-gate) · [`flail-detector`](./packages/flail-detector) · [`consensus-fix`](./packages/consensus-fix) · [`runner`](./packages/runner) |
 | **Review evidence & calibration** | [`behavior-diff`](./packages/behavior-diff) · [`gate-manifest`](./packages/gate-manifest) · [`hollow-test`](./packages/hollow-test) · [`prosecute`](./packages/prosecute) · [`review-calibration`](./packages/review-calibration) · [`model-ratchet`](./packages/model-ratchet) · [`gate-fuzzing`](./packages/gate-fuzzing) |
 | **Compounding defenses** | [`lesson-foundry`](./packages/lesson-foundry) · [`rejection-mining`](./packages/rejection-mining) · [`skill-rot`](./packages/skill-rot) |
 | **Shared foundation** | [`@adlc/cli`](./packages/cli) · [`@adlc/core`](./packages/core) |
--- a/docs/integrations/claude-code.md
+++ b/docs/integrations/claude-code.md
@@ -99,7 +99,7 @@ supplies independent model judgment with cross-lens verification.
 | flail-detection | PostToolUse | Advisory: flags repeated-error / churn loops over a bounded recent window of the transcript. |
 | gate-manifest audit | Stop | Advisory: warns only if the gate-evidence chain is broken. |
 | **rails-guard** | PreToolUse | **Enforcing**: denies structured edits (Edit/Write/MultiEdit) to frozen rail paths declared in tickets. Bash is not gated in-session (a shell can't be reliably parsed); Bash rail mutations are caught by the CI diff gate at commit time. |
-| **build-gate** | PreToolUse | **Enforcing** (issue #48): for the *active* ticket (`ADLC_TICKET` env var or `.adlc/current-ticket.json`), denies a structured edit when the ticket is high-risk (declared `risk: 'high'`, or derived from category/external-effect/identity-mutation/trust-root-touch signals) AND this session's context-fitness signal (transcript tool-call depth or byte size) is past threshold — i.e. a context-rot backstop on the riskiest builds. |
+| **build-gate** | PreToolUse | **Enforcing** (issue #48): for the *active* ticket (`ADLC_TICKET` env var or `.adlc/current-ticket.json`), denies a structured edit when the ticket is high-risk (declared `risk: 'high'`, or derived from category/external-effect/identity-mutation/trust-root-touch signals) AND this session's context-fitness signal (transcript tool-call depth or byte size) is past threshold — i.e. a context-rot backstop on the riskiest builds. Bash is not gated in-session (same reason as rails-guard) and, unlike rails, there is no CI backstop for it — see Gaps below. |
 
 All hooks no-op unless the repo is ADLC-initialized. Rail enforcement
 additionally no-ops until a ticket declares `rails`, so installing the plugin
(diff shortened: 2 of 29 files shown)


voodootikigod:
Fresh round-6 adversarial-review pass on the round-5 doc fix (8433d86):

1. **Docs claims re-verified.** The README correction ("hook does NOT shell to `adlc build-gate <id>` for its decision") checked out accurate against `plugins/adlc-claude-code/hooks/adlc-hook.mjs`'s buildgate mode line-by-line. Also checked `docs/integrations/claude-code.md` and `docs/specs/build-gate-fitness.md` for the same class of overstatement — none found.
2. **Core logic re-read in full** (`risk.mjs`, `depth-signal.mjs`, `decide.mjs`, `override.mjs`) — no correctness bugs independent of the two accepted known-limitations (Bash bypass, no-rails-declared gap). The hook's local ports match the toolkit implementations byte-for-byte, and `recordOverride()` correctly routes through `@adlc/gate-manifest`'s real chain builder.
3. **Found and fixed a real, unrelated bug**: `packages/cli/test/dispatch.test.mjs` had two hardcoded `22`-tool assertions that went stale the moment this branch registered `build-gate` as tool #23 (`TOOLS.length` assertion + a `/Tools \(22\)/` regex). Fixed in `fb1f38b`, and made the count assertion derive from `TOOLS.length` so it can't go stale again on a future add/remove.

All suites green: build-gate 67/67, cli 20/20 (was failing pre-fix), claude-code hooks 69/69, full repo suite exit 0.

**Recommend removing `[NEEDS ATTENTION]`** — round 5's doc fix holds up, no new logic bugs, the two documented Bash/no-rails gaps remain accepted scope per prior discussion.

Issue #159, opened 2026-07-12
rails have no START lifecycle: enforced from authoring, not from build-start (mirror of #141)


voodootikigod:
## Context
Surfaced building both T39 and T40 — the **mirror** of #141 (rails have no *completion* lifecycle). Both tickets shipped with `rails: []` because declaring rails at **authoring** freezes those paths in the rails-guard union *immediately* — before the build even starts — blocking unrelated PRs that touch them. So the honest move was to declare no rails and note "declare at P3," which defeats the point of authoring rails up front.

Rails have no **start** lifecycle: they are enforced from the moment a ticket exists, not from when its build begins.

## Proposal
Enforce a ticket's rails only within its **active build window** — from P3-start (rails declared/frozen) to completion — not from authoring. Options: a per-ticket `railsActiveFrom`/status the rails union respects, or deriving "active" from a gate-manifest P3 entry. Complements #141 (rails auto-expire on completion) — together they give rails a full start→end lifecycle.

## Relationship
- #141: completion side (rails expire when `completed: true`).
- This: start side (rails don't freeze until the build begins).

## Acceptance
- A ticket can declare its rails at authoring WITHOUT freezing those paths for unrelated PRs before its build starts.
- Once the build starts (P3), the rails freeze as today; on completion they expire (#141).

PR #266, opened 2026-07-22
fix(ci): recognize the bot author gh reports, so the drift tracker updates (#265)

### LINKED ISSUE #265: ceremony-drift opens a duplicate tracker on every push: MANAGED_AUTHORS never matches gh's bot login
The ceremony-drift job promises a single self-maintaining tracking issue ("This issue closes on its own once the set is empty; edits to the body are overwritten"). It instead opened 12 issues in two days (#230, #245, #246, #247, #250, #254, #256, #257, #259, #261, #262, #264) — one per push to `main`.

## Root cause

`scripts/ceremony-drift.mjs`:

```js
const MANAGED_AUTHORS = (process.env.ADLC_DRIFT_AUTHORS ?? "github-actions[bot]")
```

`github-actions[bot]` is the **REST** actor name. `gh issue list --json author` returns the **GraphQL** actor, which gh renders as `app/github-actions`:

```
$ gh issue view 264 --json author
{"author":{"is_bot":true,"login":"app/github-actions"}}
```

`findExistingIssue` passes `authors: MANAGED_AUTHORS` to `selectTrackingIssue` on **both** lookup paths — the labeled fast path (`requireMarker: false`) and the unlabeled sweep. The author filter therefore drops the genuine tracker in both, `findExistingIssue` returns `null`, and `decideAction` takes the `open` branch on every run.

Run log for the most recent push confirms it:

```
ceremony-drift: 11 ticket(s) awaiting the completion ceremony
ceremony-drift: opened https://github.com/voodootikigod/adlc/issues/264
```

Every run exits 0, so nothing surfaced the malfunction. The `ambiguous tracking issue` guard never fired either — it counts candidates *after* the author filter, which is always empty.

## Why the tests did not catch it

`scripts/test/ceremony-drift.test.mjs` fabricates the login it asserts against:

```js
const issues = [{ number: 9, body: `${MARKER} real`, author: { login: "github-actions[bot]" } }];
assert.equal(selectTrackingIssue(issues, { authors: ["github-actions[bot]"] }).number, 9);
```

The fixture and the production default share the same wrong string, so the test passes on a value `gh` never emits. `selectTrackingIssue` is correct in isolation; the defect is entirely in the constant it is fed, and no test covers the constant against a real `gh` payload.

## Fix

1. Normalize actor logins before comparing — strip the `app/` prefix and the `[bot]` suffix so `app/github-actions`, `github-actions[bot]`, and `github-actions` all compare equal. Both gh renderings and the REST name then match, and the fix survives a future gh formatting change.
2. Replace the fabricated-login test fixtures with the real recorded `gh --json author` payload shape, and add a case asserting `app/github-actions` is accepted by the default `MANAGED_AUTHORS`. A test that only exercises a hand-written login cannot catch a constant that is wrong about the outside world.

## Cleanup already done

The 11 duplicates are closed; #230 is retained as the canonical tracker (label and marker intact), so the corrected lookup will find and update it in place rather than opening a 13th.

voodootikigod:
Fixes #265. Ticket T58.

## The defect

`ceremony-drift` promises one self-maintaining tracking issue. It opened 12 in two days — one per push to `main`.

`MANAGED_AUTHORS` defaulted to `github-actions[bot]`, the **REST** actor name. `gh issue list --json author` returns the **GraphQL** actor, which gh renders `app/github-actions`:

```
$ gh issue view 264 --json author
{"author":{"is_bot":true,"login":"app/github-actions"}}
```

`findExistingIssue` passes that filter to `selectTrackingIssue` on **both** lookup paths — the labeled fast path and the unlabeled sweep. Neither could match, `findExistingIssue` returned `null`, and `decideAction` took the `open` branch every run:

```
ceremony-drift: 11 ticket(s) awaiting the completion ceremony
ceremony-drift: opened https://github.com/voodootikigod/adlc/issues/264
```

Every run exited 0, so nothing surfaced the malfunction. The `ambiguous tracking issue` guard never fired either — it counts candidates *after* the author filter, which was always empty.

## Why the tests did not catch it

They fabricated the login they asserted against:

```js
const issues = [{ number: 9, body: `${MARKER} real`, author: { login: "github-actions[bot]" } }];
assert.equal(selectTrackingIssue(issues, { authors: ["github-actions[bot]"] }).number, 9);
```

The fixture and the production default shared the same wrong string, so they agreed with each other while both disagreed with `gh`. `ceremony-drift-exit.test.mjs` had the same fabricated login in its `gh` stub — so even the end-to-end `assert.doesNotMatch(r.calls, /issue create/)` ("never a duplicate") passed while production duplicated on every push. A stub is only evidence if it lies the way the outside world does.

## The fix

- `normalizeActorLogin` strips the `app/` prefix and `[bot]` suffix and lowercases, so every rendering of one actor is one identity. Normalizing at the comparison rather than correcting the constant to today's rendering is what makes the next gh formatting change a non-event.
- **Bot-ness is required alongside the name.** Stripping `[bot]` alone would let a human account literally named `github-actions` normalize onto the managed identity and have its issue seized by a job holding `issues: write` — a hole the exact-match comparison did not have. `is_bot` absent means not provably a bot, which fails closed.
- Fixtures replaced with the recorded `gh --json author` payload, in both the unit and the end-to-end suite.

## Evidence the tests are load-bearing

Reintroducing the exact-string comparison:

```
not ok 8  - an UNLABELED tracker authored by the bot is adopted, relabeled, and updated
not ok 35 - the default managed-author set accepts the author gh actually reports
not ok 36 - every rendering of the same bot actor is one identity
not ok 38 - a recognized tracker updates in place instead of opening a duplicate
# pass 80  # fail 6
```

Removing the bot-ness requirement:

```
not ok 21 - a human account whose name normalizes to the bot is still rejected
# pass 69  # fail 1
```

Full suite: `46/46 segments passed`, 138 tests, 0 failures.

## Cleanup already performed

The 11 duplicates (#245, #246, #247, #250, #254, #256, #257, #259, #261, #262, #264) are closed as not-planned with a pointer comment. **#230 is retained as the canonical tracker**, label and marker intact, so the corrected lookup finds and updates it in place rather than opening a 13th.

## Follow-up (not in this PR)

The ceremony itself — completing the 11 shipped rail-freezing tickets — lands separately on the protected-base path, per the precedent in #199, #221, #227.

## Test plan

- [x] `node --test scripts/test/ceremony-drift.test.mjs scripts/test/ceremony-drift-exit.test.mjs` — 86 pass
- [x] `npm test` — 46/46 segments
- [x] Mutation-verified: both the defect and the escalation guard are caught
- [ ] After merge: confirm the next `ceremony-drift` run logs `updated issue #230` rather than `opened`

### DIFF
--- a/.adlc/tickets.json
+++ b/.adlc/tickets.json
@@ -1268,6 +1268,24 @@
       "title": "A version-only manifest edit must not read as a rail edit (#228)"
     },
     {
+      "body": "Fixes voodootikigod/adlc#244. changedFiles(base) in packages/core/lib/git.mjs only diffs base-vs-working-tree, so staging a rail violation then reverting the working copy evades rails-guard. Union git diff <base> and git diff --cached <base> changed-file sets. Verification: node --test packages/core/test/*.test.mjs packages/rails-guard/test/*.test.mjs. allow-suppression: none",
+      "budget": "mid",
+      "category": "bug",
+      "duration": 1,
+      "edges": [],
+      "id": "T56",
+      "rails": [
+        "packages/rails-guard/test/version-only.test.mjs",
+        "packages/rails-guard/test/version-only-e2e.test.mjs"
+      ],
+      "scope": [
+        "packages/core/lib/git.mjs",
+        "packages/core/test/*.test.mjs",
+        "packages/rails-guard/**"
+      ],
+      "title": "rails-guard: union staged and working-tree diffs so a staged-then-reverted rail edit fails the gate"
+    },
+    {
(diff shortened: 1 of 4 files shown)


voodootikigod:
Review rounds resolved — pushed as a7f1502.

**Adversarial review (medium, compatibility)** — requiring bot-ness on *every* acceptance route broke the documented `ADLC_DRIFT_AUTHORS` override. A dedicated machine user is reported with `is_bot:false`, so `ADLC_DRIFT_AUTHORS=release-machine` stopped matching and recreated #265 under that configuration. Fixed by making the two routes deliberately asymmetric: an exact case-insensitive match against a configured entry is explicit operator intent and needs no further evidence; only the *normalized* route widens the match beyond what was configured, and only there is bot evidence required. Re-review: **APPROVE, no material findings.**

**Mutation gate** — two surviving mutants, both genuine hollow spots:

- `bool-flip` on `if (!login) return false` — nothing noticed when the no-login guard stopped guarding. An author carrying no login carries no authorization evidence and must be rejected, not treated as anonymous-and-therefore-fine.
- `logic-swap` on `isBotFormLogin` (`||` → `&&`) — every fixture set `is_bot: true`, which short-circuits the check before the login shape is ever consulted, so the function was entirely unexercised. New cases match only through normalization and omit `is_bot`, the one path where the shape is load-bearing.

`mutation-gate`: **12/12 mutants killed**. All 20 CI checks pass. Local: 46/46 segments.

**Process note for #159.** Authoring this ticket's tests required narrowing T58's rails and re-freezing them three times — once to write the tests at all, then once per review round. Rails are enforced from *authoring* rather than from build-start, so declaring a test file as a rail blocks writing that very test. Each cycle needed `--authorize` on a rail-narrowing, which is recorded, so nothing was bypassed — but it is friction pointing straight at #159, and worth citing there as a concrete instance.

PR #285, opened 2026-07-22
chore(tickets): complete 12 shipped rail-freezing tickets (ceremony)

voodootikigod:
Clears the drift tracked by #230. Follow-up to #266.

## What this is

Twelve tickets shipped while freezing rails. A shipped ticket must be marked `completed` for its rails to expire (T36); nothing did that, so stale rails kept freezing unrelated future work — the failure that blocked PR #196 and motivated #198/#200.

## Evidence

Each ticket was confirmed against **the merge that shipped it**, not against the reporter's inference. The issue body is explicit that this matters: "all declared scope globs resolve to tracked files" is equally true of an active ticket editing existing files, so that signal cannot distinguish finished from in-progress.

| Ticket | Shipped in |
|---|---|
| T-01KXPD8KJ9H6M6DFA83Y82A1Z1 | #196 active-ticket pointer contract |
| T-01KY00D1KX0HDE3521KTFC9T26 | #214 release artifact gate |
| T46 | #181 native Codex integration (+#192) |
| T48 | #200 / #207 ticket-prune + drift reporter |
| T50 | #201 Stop-time review nudge (Codex) |
| T54 | #201 P5 prosecutor fan-out (Codex) |
| T56 | #255 staged/working-tree diff union |
| T57 | #194 → #258 adlc-init on a fresh repo |
| T58 | #266 ceremony-drift bot-author fix |
| T7, T8, T9 | #33 ticket-sync (+#51 for T9) |

Every declared rail was additionally verified present on `main`.

T58 is included because it shipped in #266 during this same sweep — completing it now avoids handing the next run a fresh drift entry.

## Safety

- The four genuinely-active tickets (**T13, T38, T51, T53**) are untouched, and none declares a rail over any of the completed work — so no in-flight build loses a rail.
- No active-ticket pointer exists in this checkout, and the review was run from a local checkout rather than relying on CI's (which structurally cannot see one — `.adlc/current-ticket.json` is gitignored).
- Diff is **purely additive**: one `completed: true` per ticket, 12 insertions, no existing field changed.

## Result

```
$ adlc ticket-prune --base-ref origin/main
Active tickets (4):
  - T13  - T38  - T51  - T53
```

The "Shipped but still freezing rails" section is gone entirely. Once this lands, #230 should close itself on the next `ceremony-drift` run — which is now the fixed reporter, verified in production on the #266 merge:

```
ceremony-drift: updated issue #230
```

(first run in two days that updated rather than opened a duplicate).

## Merge path

[...]

Issue #141, opened 2026-07-11
Trust-root change ceremony: a CI-recognized authorized path instead of routine admin-override of a red check


voodootikigod:
## Context

Surfaced dogfooding #104 (PR #135). Landing any legitimate change to a trust root (`rails-guard-ci.mjs`, `docs/ci/rails-guard.yml`, the hash fixture, `.adlc/tickets.json`) means:

1. The `rails-guard` CI check goes **red by design** — the gate is self-protecting and denies its own modification.
2. A human clicks **admin-merge** to override the red required check.

This works and is the documented protected-base path, but the audit trail is just *"someone used `--admin`."* There's no explicit, logged signal that *this specific trust-root change was reviewed and authorized* — and, worse, it **trains operators to override a red gate**, which is a bad reflex to build around security controls.

## Proposal

Add a **CI-recognized authorized-change ceremony** so the green/red signal stays meaningful:

- A CODEOWNERS-gated label (e.g. `trust-root-change`) that, when applied by an authorized reviewer, makes `rails-guard-ci` treat the self-modification as **authorized** (exit 0) — while still **recording the change to the gate-manifest** with who authorized it.
- Equivalent to the existing `ADLC_RAILS_BYPASS=1` audited path, but surfaced as an explicit, reviewer-gated PR state instead of a post-hoc admin click.
- Net: legitimate trust-root PRs go **green** (authorized), unauthorized self-modifications stay **red** (denied), and the authorization is explicit and logged rather than an opaque bypass.

## Open questions

- Binding authorization to CODEOWNERS approval vs a signed manifest entry (avoid exposing a signing key to PR CI — see the #104/T36 discussion where manifest-signature approaches were rejected as a security regression for exactly this reason).
- Ensuring the label can't be self-applied by the PR author.

## Acceptance sketch

- A trust-root PR with the authorized label + required approval passes `rails-guard` (recorded to manifest); without it, still fails.
- No routine admin-override of a red required check for authorized trust-root changes.

## Evidence

- #104 / PR #135 (admin-merged over the by-design red `rails-guard` check); T36 manifest-signature rejection rationale.

voodootikigod:
## Grooming: this is the anchor of milestone `B2` — build it first

Four other issues all edit `scripts/rails-guard-ci.mjs` and/or `docs/ci/rails-guard.yml`, which are self-protecting trust roots that turn the `rails-guard` check red by design:

[...]

PR #287, opened 2026-07-22
spec(adlc): shape the rail START lifecycle (#159) and trust-root ceremony (#141)

voodootikigod:
Shaping only — two specs and two tickets, no implementation. Closes nothing yet; T59 and T60 carry the work.

## Why these are one PR

They are the two ends of the same lifecycle, and neither is coherent alone.

- **#159 — the START side.** Rails freeze from the moment a ticket exists. T36 (#130) gave rails an END; nothing gives them a beginning.
- **#141 — the mechanism that makes the drift unclearable.** `isCompletionAnnotationOnly` (rails-guard-ci.mjs:147) bails on `baseRails.length > 0`, so completing a *railed* ticket — the one operation that clears rail drift — cannot land on the normal path.

Together: rails accumulate from authoring, and can only be cleared by overriding a red required check. Friction on clearing drift is structurally higher than friction on creating it, so the union ratchets upward.

**The measurement that makes this concrete.** The active rail union on `main` at 1.6.0 is **30 paths, and all 30 belong to tickets that already shipped** (the twelve in #285). Not one frozen path corresponds to a build in flight. That is not an accident — it is the steady state this design produces.

## Correction to #159

Its Relationship section cites #141 as the completion side. That is stale: completion shipped as **T36 / #130**. #141 is the mechanism problem, not the lifecycle mirror. Worth fixing on the issue.

## Gates run, findings folded back in

| Gate | Result |
|---|---|
| `spec-lint` | 34/34 acceptance criteria verified, 0 wishes |
| `premortem` (`--prompt-only`) | 7 findings, **2 blocking** |
| `coldstart` (`--prompt-only`) | 3 gaps, **1 blocking** |

Findings were folded into the specs, not appended as caveats.

### Blocking — premortem

**Shipping #159 before #285 silently expires all 30 frozen paths at once.** A mass unfreeze arriving as a side effect of a lifecycle change rather than a reviewed ceremony, quietly mooting #285. Absent `started` must therefore *not* mean "never started" for pre-existing tickets. The grandfather clause is required regardless of merge order — correctness must not depend on it.

**A `trust-root-authorized` check run is forgeable.** This was my first design, and it is wrong: a same-repo PR's `GITHUB_TOKEN` carries `checks: write` by default, so a PR can create a check run with that exact name on its own head SHA and authorize itself. A check-run name is not an authorization; it is a string any writer can produce. The check run is now explicitly **a cache, never the authority** — `rails-guard-ci.mjs` re-queries the label applier and approving reviewer and re-derives the decision.

### Blocking — coldstart

**This repo has no CODEOWNERS file** (`ls CODEOWNERS .github/CODEOWNERS` → nothing), so #141's authorization bar is unimplementable until one exists. Authoring it declares *who may approve trust-root changes* — a governance decision for you, not something a build agent can infer. Recorded as a blocking input on T60, and **the one thing I need from you before T60 can start.**

### Other findings folded in

- Build step 7 (automatic P3 transition) was a requirement with **no acceptance criterion** — it would have shipped the flag while leaving it a remembered step. Now AC16, and `packages/build-gate/**` added to T59's scope, since that is where the transition lives.
- The prescribed `formatVersion` bump **collides with existing code**: rails-guard-ci.mjs:348 hardcodes `base===0 && head===1` and :377 denies any other version change, so bumping to `2` would deny every PR. Now AC17, and the spec prefers a per-ticket provenance signal so store version keeps meaning "layout", not "policy epoch".
- Authorization bar tightened from `write` to CODEOWNERS (write covers every contributor with push access).
- Fork PRs refused explicitly — a fork's head SHA does not exist in the base repo, so nothing can bind to it.
- Start+complete recovery in one PR matches *neither* narrow exemption. Left as a decided choice with an AC either way; the failure mode was it being undefined.

## The security argument for `started` (the crux of T59)

`started: true` only ever **adds** freeze, so granting it takes privilege away from the setter — an ordinary PR may set it. Un-starting is already denied by `assertBaseTicketContractsPreserved`, so the transition is one-way for free. And no new unfreeze privilege is created: the union only shrinks for tickets that never started, and a never-started ticket is *exactly equivalent to today's `rails: []`*, which any author may already write. The floor is unchanged.

Note the deliberate asymmetry with completion: `isCompletionAnnotationOnly` bails on railed tickets because completing one *unfreezes*. Starting one *freezes*, so the railed case is precisely the case that must be allowed. Getting this backwards inverts the security property.

## Test plan

- [x] `adlc spec-lint` both specs — 34/34 verified, 0 wishes
- [x] `adlc premortem --prompt-only` both — findings folded in
- [x] `adlc coldstart --prompt-only` T59/T60 — gaps folded in
- [x] `adlc ticket create/update` — T59, T60 authored; scope-widening authorized and recorded
- [ ] Implementation lands under T59 and T60 separately
- [ ] **Blocked on you:** CODEOWNERS ownership decision before T60 starts

### DIFF
--- a/.adlc/specs/rails-start-lifecycle.md
+++ b/.adlc/specs/rails-start-lifecycle.md
@@ -1,0 +1,296 @@
+# Rails need a START lifecycle: freeze from build-start, not from authoring
+
+Closes #159.
+
+## Problem
+
+A ticket's rails freeze their paths for **every** PR from the moment the ticket
+exists in the store. `scripts/rails-guard-ci.mjs:329-335` builds the union from
+every base ticket that is not completed:
+
+```js
+const rails = [];
+for (const t of baseTickets) {
+  if (t.completed === true) continue; // done → auto-expire (T36)
+  for (const r of t.rails ?? []) rails.push(r);
+}
+```
+
+There is no notion of *when the build began*. T36 (#130) gave rails an END —
+`completed: true` expires them. Nothing gives them a START. The window is
+therefore `[authoring, completion]` when it should be `[build-start, completion]`.
+
+(#159's "Relationship" section says the completion side is #141. That is a stale
+cross-reference: the completion side shipped as **T36 / #130**. #141 is the
+trust-root ceremony, which is the *mechanism* problem, not the lifecycle mirror.
+Worth correcting on the issue.)
+
+### Harm A — a ticket cannot author its own rails
+
+Declaring a test file as a rail blocks writing that test file. The rail is
+supposed to freeze a test *once it exists*; instead it freezes the path before
+anything is there.
+
+Observed building **T58** (the ceremony-drift bot-author fix) on 2026-07-22. The
+ticket declared `rails: ["scripts/test/ceremony-drift.test.mjs"]` at authoring,
+and the first attempt to write that test was denied:
+
+```
+scripts/test/ceremony-drift.test.mjs is a frozen rail declared by ticket T58
+(rails: "scripts/test/ceremony-drift.test.mjs"). Edits to frozen rails are
+blocked during build.
+```
+
+The workaround was to narrow the rails to `[]`, author the test, then re-freeze
+— and to repeat that cycle on **every review round** that added a test (an
+adversarial-review finding, then two surviving mutants). Three narrow/re-freeze
+cycles for one ticket, each requiring `--authorize` on a `rail-narrowing`.
+
+Nothing was bypassed and every cycle is recorded, so this is friction rather than
+a hole. But the recorded history now shows a ticket that repeatedly un-froze its
+own rails, which is indistinguishable in the log from the thing rails exist to
+prevent. The audit trail is *noisier for having followed the rules*.
+
+### Harm B — the honest response is to declare no rails
+
+#159 reports that T39 and T40 both shipped with `rails: []` and a note to
+"declare at P3", because declaring up front would freeze those paths for
+unrelated PRs before the build started. That defeats the point of authoring
+rails with the ticket, and it is the *documented* workaround, not a mistake.
+
+The in-session hook states the current behavior plainly
+(`plugins/adlc-claude-code/hooks/adlc-hook.mjs:250`):
+
+```js
+const status = ticket.completed === true ? 'completed'
+  : rails.length > 0 ? 'rail protection auto-active' : 'no rails declared';
+```
+
+`auto-active` — on declaration.
+
+### Why this is worth fixing now
+
+Measured against the store on `main` at 1.6.0: the active rail union is **30
+paths**, and **every one of them belongs to a ticket that has already shipped**
+(the twelve in PR #285). Not one frozen path corresponds to a build in flight.
+
+That is the steady state this design produces. Rails accumulate from authoring,
+expire only through an admin ceremony, and the union therefore trends toward
+"everything ever declared" rather than "what is being built right now." #159 and
+#141 are the two ends of the same lifecycle, and the union is broken at both.
+
+## Build
+
+1. **Add a `started` field and gate the union on it.** In the
+   `rails-guard-ci.mjs` union, skip tickets that have not started:
+
+   ```js
+   if (t.completed === true) continue; // T36 — window closed
+   if (t.started !== true) continue;   // NEW  — window not open
+   ```
+
+   Strict boolean `true` only, matching T36's fail-closed treatment of
+   `completed` (not `"true"`, not truthy).
+
+2. **Make `started` self-service, and irreversible.** This is the trust
+   asymmetry that makes the change safe, and it must be argued explicitly in the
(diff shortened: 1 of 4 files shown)


voodootikigod:
Reworked T60's authorization model, and added CODEOWNERS.

## The CODEOWNERS gap was deeper than a missing file

I flagged "no CODEOWNERS file" as a governance decision for the maintainer. Checking who could actually be in it showed the real problem:

```
$ gh api repos/voodootikigod/adlc/collaborators --jq ".[].login"
voodootikigod          # the only one
$ git log -200 --format="%an" | sort -u
Chris Williams
```

One collaborator. And **GitHub structurally forbids approving your own pull request** — not a policy toggle, it cannot be done. So two clauses in my draft were unsatisfiable:

- *"the label applier … is not the PR author"* — nobody else to apply it
- *"an `APPROVED` review … not the author"* — impossible

**As specified, this ceremony would have authorized nothing** — a red check with *no* path through it, strictly worse than the admin override it replaces. That is a flaw in my spec, not a config gap. Coldstart caught that CODEOWNERS was missing; it did not catch that the rule was unsatisfiable, because I asked whether the file existed and never asked who could possibly be in it.

## What changed

Separation of duties is genuinely unavailable on a solo repo, and no design recovers it — a second identity the same person controls is theatre, not a control. The spec now says so plainly rather than implying a property it cannot deliver.

What survives, and is the actual justification for building this at all:

- the required check stays **green**, so operators are not trained to override a red gate — the primary harm in #141
- authorization is **revision-bound**; pushing invalidates it, which an admin click has no notion of
- it is **recorded** with actor, SHA and reason, not an opaque "someone used `--admin`"
- it is **narrowly scoped** to one trust-root change, not "bypass the gate"

Strictness is now **derived from the eligible-CODEOWNER count**:

| Mode | Rule |
|---|---|
| **single-owner** | Owner self-authorizes by label. No review required (none can exist). Manifest records `mode: single-owner`. |
| **multi-owner** | Full separation of duties: distinct applier, approving CODEOWNER review, neither the author. |

Two properties make this safe rather than a permanent excuse:

- **AC18 — it only ratchets up.** Mode is computed from CODEOWNERS at evaluation time. Adding a second owner engages the strict rule automatically — no flag, no config, no migration — and nothing can select the weaker mode while a stronger one is available.
- **AC19 — single-owner evidence is honest.** The manifest entry carries `mode: single-owner` and no reviewer field, so the audit trail never implies a second pair of eyes that did not exist. An entry claiming a reviewer in single-owner mode is a defect.

## CODEOWNERS

Added, covering the five trust-root paths (the gate, its test, the CI definition, the hash fixture pinning that definition, and the ticket store). Its header states the single-owner consequences explicitly so nobody later mistakes the file for a separation-of-duties control.

It is worth having independently of T60 — an unowned trust root is a gap whether or not anything enforces ownership.

`spec-lint`: 19/19 verified, 0 wishes (was 17).

voodootikigod:
Rebased onto current `main` — **all checks green**, `mergeStateStatus: CLEAN`.

## The failure

`mutation-gate` was failing for the reason this PR itself surfaced: adding `CODEOWNERS` was misclassified as mutable source, which forced the FULL-suite slow path, whose baseline is broken (#288).

```
mutation-gate: 1 file(s) have no known fast test target — falling back to the FULL suite:
  CODEOWNERS
error: baseline suite is not green (exit 1)
```

That was fixed in **#289**, now merged. This branch was simply running against pre-fix `main`; nothing in the shaping change was wrong.

Both halves of #289 are doing work here:

```
mutation-gate: 4 changed file(s) are not a mutable source type, so this gate does not cover them:
  .adlc/specs/rails-start-lifecycle.md
  .adlc/specs/trust-root-ceremony.md
  .adlc/tickets.json
  CODEOWNERS
mutation-gate: no source files changed — nothing to mutate.
```

`CODEOWNERS` is correctly non-source so the slow path is never taken, **and** the skip is reported rather than silent — you can tell the gate looked and found nothing, versus did not look.

## Conflict resolution worth reviewing

`.adlc/tickets.json` conflicted: `main` had gained the twelve ceremony completions (#285) and T61 (#289).

I resolved to **main's version** and re-created T59/T60 through `adlc ticket create`, rather than hand-merging. Hand-editing the store risks desynchronising `planHash`/`storeHash` from the ticket contents, and those hashes are what the gates trust. Re-creating means they are computed, not guessed.

Verified after the rebase:

- T59 and T60 present, T59's dependency edge to T60 intact
- `spec-lint`: 17/17 and 19/19 criteria verified, 0 wishes
- The full `ticket-store-platform` matrix passes on all three OSes × Node 18/20/22 — the check that actually exercises a rewritten store

Diff is 4 files, unchanged in substance from before the rebase.

## Still blocked on you

**T60 needs no further input** — the CODEOWNERS ownership question is resolved in the spec (owner-count-conditional authorization). **T59 was never blocked.** Both tickets are executable.

Issue #288, opened 2026-07-22
hollow-test discards the baseline suite output on failure, making a red slow-path baseline undiagnosable


voodootikigod:
Split out from PR #287. Distinct from the source-classification bug fixed there.

## Symptom

The `mutation-gate` slow path (FULL-suite fallback) fails in CI with a baseline that is green everywhere else:

```
mutation-gate: 1 file(s) have no known fast test target — falling back to the FULL suite:
  CODEOWNERS
mutation-gate: capping budget at 3 mutant(s) for the slow path (requested 12)
error: baseline suite is not green (exit 1) — cannot measure mutation kill; fix the suite / --test-cmd first
```

On the same commit, `test (18)`, `test (20)` and `test (22)` all pass, and `node scripts/run-tests.mjs` reports `46/46 segments passed`.

## What I ruled out

- **Not a timeout.** `hollow-test.mjs:133` distinguishes the two; the message says `exit 1`, not `timed out`.
- **Not `npm ci` vs `npm install` at the lockfile level.** The `test` job uses `npm install`, `mutation-gate` uses `npm ci`, but `npm ci` succeeded — it fails outright on a lock/manifest mismatch.
- **Not a genuine suite failure under the exact lockfile.** A clean `git clone` + `npm ci` + `node scripts/run-tests.mjs` is green: 46/46 segments, 138 tests, 0 failures.
- **Not missing platform-specific optional deps.** `package-lock.json` carries the full `@esbuild/linux-*` set alongside `darwin-*`.

## Why it is not diagnosable

`packages/hollow-test/bin/hollow-test.mjs`:

```js
const baseline = runTest(testCmd, timeoutMs, cwd);
if (baseline.status !== 0) {
  const reason = baseline.timedOut ? "timed out" : `exit ${baseline.status}`;
  opError(`baseline suite is not green (${reason}) — cannot measure mutation kill; fix the suite / --test-cmd first`);
}
```

The baseline's stdout and stderr are captured and then **dropped**. All that survives is the exit code, so the one artifact that would identify the failing segment never reaches the log. The message says "fix the suite" without saying what broke — and the suite is green under every other invocation, so there is nothing to act on.

## Proposed fix

1. **Make the failure legible (the actual bug here).** On a non-zero baseline, emit the captured stdout/stderr — truncated if large. A gate that refuses to run must say why; otherwise its failure mode is indistinguishable from a flake and the cost lands on whoever next trips it.
[...]

PR #289, opened 2026-07-22
fix(hollow-test): one source predicate, JS-only, with declared overrides

voodootikigod:
Ticket T61. Unblocks #287. Surfaced #288.

## The bug

`hollowTestWouldMutate` (scripts/mutation-gate.mjs:117) decided what counted as mutable source by **excluding** a fixed set of non-source extensions:

```js
const HOLLOW_TEST_EXCLUDE_EXT = /\.(?:md|json|yml|yaml|lock|txt|toml|snap)$/i;
```

That answers the wrong question — "is this one of the non-source extensions I happened to think of?" Anything with **no extension** matches nothing and falls through as code:

```
MUTATES CODEOWNERS      MUTATES Makefile      MUTATES .nvmrc
MUTATES LICENSE         MUTATES .gitignore
MUTATES Dockerfile
```

Adding a `CODEOWNERS` file on #287 was enough to strand the gate on the FULL-suite slow path:

```
mutation-gate: 1 file(s) have no known fast test target — falling back to the FULL suite:
  CODEOWNERS
```

An exclusion list is **unbounded by construction**. Every extensionless file anyone adds later is a fresh false positive, and each is discovered the same expensive way: a red required gate on an unrelated PR.

`hollow-test` can only mutate JS-family source, so enumerate that instead. This also resolves a disagreement inside the gate itself — the workflow step already filters `'*.mjs' '*.js' '*.cjs' 'packages/**' 'plugins/**'`, so the script and its own workflow held different definitions of source.

## Tests assert both directions

The second direction matters more than the first. A fix that traded false positives for **silently skipping real source** would be far worse than the bug — the coverage gate would pass by not looking.

- extensionless files (`CODEOWNERS`, `LICENSE`, `Dockerfile`, `Makefile`, `NOTICE`, `Procfile`) → not source
- dotfiles (`.gitignore`, `.nvmrc`, `.npmrc`, `.editorconfig`, `.gitattributes`) → not source
- every JS-family extension (`.mjs`, `.cjs`, `.js`, nested under `packages/` and `plugins/`) → still source

Verified against the real #287 diff:

```
classify(["CODEOWNERS", ".adlc/specs/a.md", ".adlc/tickets.json"])
→ { kind: "nothing" }
```

## Filed separately: #288

Diagnosing this exposed a second, independent defect. The slow-path baseline is red in CI while green everywhere else, and it is **undiagnosable** because `hollow-test` captures the baseline's stdout/stderr and then drops them, keeping only the exit code.

I ruled out a timeout (the message says `exit 1`), an `npm ci`/`npm install` lockfile mismatch (`npm ci` succeeded), a genuine failure under the exact lockfile (clean clone + `npm ci` + full suite → 46/46 green), and missing platform optional deps (the full `@esbuild/linux-*` set is in the lock). Past that point there is nothing to work with, which is the actual bug.

Worth noting: the slow path appears to have **never run in CI** since the gate landed in #251 — every PR since touched only files with known fast test targets. It is a latent break one unusual file away from blocking any PR. This fix makes that branch *rarer*, which makes it more likely to rot unnoticed, so #288 asks for the output to be made legible and the branch deliberately exercised.

## Test plan

- [x] `node --test scripts/test/mutation-gate.test.mjs` — 22 pass, 0 fail; the two new cases fail against the pre-fix classifier
- [x] `node scripts/mutation-gate.mjs --base main` on this change — 1/1 mutants killed
- [x] Full suite
- [ ] After merge: rebase #287 and confirm `mutation-gate` goes green

### DIFF
--- a/.adlc/config.json
+++ b/.adlc/config.json
@@ -1,0 +1,9 @@
+{
+  "version": 1,
+  "securityMode": "unsigned-fallback",
+  "harnesses": {
+    "cursor": {
+      "railEnforcement": "auto"
+    }
+  }
+}
--- a/.adlc/config.json
+++ b/.adlc/config.json
@@ -1,9 +1,0 @@
-{
-  "version": 1,
-  "securityMode": "unsigned-fallback",
-  "harnesses": {
-    "cursor": {
-      "railEnforcement": "auto"
-    }
-  }
-}
--- a/.adlc/tickets.json
+++ b/.adlc/tickets.json
@@ -1364,7 +1364,10 @@
         "scripts/mutation-gate.mjs",
         "scripts/test/mutation-gate.test.mjs",
         "packages/hollow-test/lib/targets.mjs",
-        "packages/hollow-test/test/**"
+        "packages/hollow-test/test/**",
+        "packages/hollow-test/bin/hollow-test.mjs",
+        "packages/hollow-test/README.md",
+        "docs/tools/hollow-test.md"
       ],
       "title": "mutation-gate: classify source by an include-list, not by excluding known non-source extensions"
     },
--- a/.adlc/tickets.json
+++ b/.adlc/tickets.json
@@ -1353,6 +1353,22 @@
       "title": "Schema bedrock: definition-driven validator + generated JSON Schemas"
     },
     {
+      "body": "hollowTestWouldMutate (scripts/mutation-gate.mjs:117) decides what counts as mutable source by EXCLUDING a fixed set of non-source extensions (md|json|yml|yaml|lock|txt|toml|snap). Any path with no extension therefore falls through and is treated as source: CODEOWNERS, LICENSE, Dockerfile, Makefile, .gitignore, .nvmrc all return true. A PR touching any of them has no fast test target, so the gate falls back to the FULL suite on the slow path. Observed on PR #287: adding CODEOWNERS produced 'mutation-gate: 1 file(s) have no known fast test target - falling back to the FULL suite: CODEOWNERS'. hollow-test can only mutate JS-family source, and the gate's own workflow step already filters '*.mjs' '*.js' '*.cjs' 'packages/**' 'plugins/**' - so the script and the workflow disagree about what source is. Invert to an include-list of JS-family extensions. Verification: node --test scripts/test/mutation-gate.test.mjs. allow-suppression: none ADVERSARIAL-REVIEW (HIGH, test-integrity): the first fix narrowed to mjs|cjs|js only, which SILENTLY SKIPPED the repo's 49 tracked non-test .ts/.tsx source files — hollow-test's own filterTargetFiles accepts them, so classify() returned kind:'nothing' and the required gate exited 0 without running. That trades a false positive for a false negative, strictly the worse direction. Root cause of BOTH bugs is the same predicate duplicated in packages/hollow-test/lib/targets.mjs:10 and scripts/mutation-gate.mjs — they drifted. Fix is one exported predicate shared by both, include-listing the JS and TS families, with a cross-contract test proving the two selectors agree.",
+      "budget": "mid",
+      "category": "bug",
+      "duration": 1,
+      "edges": [],
+      "id": "T61",
+      "rails": [],
+      "scope": [
+        "scripts/mutation-gate.mjs",
+        "scripts/test/mutation-gate.test.mjs",
+        "packages/hollow-test/lib/targets.mjs",
+        "packages/hollow-test/test/**"
+      ],
+      "title": "mutation-gate: classify source by an include-list, not by excluding known non-source extensions"
+    },
+    {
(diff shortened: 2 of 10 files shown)


voodootikigod:
Adversarial review caught a HIGH finding on the first commit, and it was the exact failure that commit's own message claimed to guard against. Fixed in eeba633.

## The finding

Narrowing to `mjs|cjs|js` silently dropped this repository's **49 tracked non-test `.ts`/`.tsx` source files**. `hollow-test`'s own `filterTargetFiles` accepts them, so `classify()` returned `kind: "nothing"` and the required gate **exited 0 without running**.

That trades a false positive (extra slow-path runs) for a false negative (source silently unmutated) — strictly the worse direction. I wrote *"a fix that traded false positives for silently skipping real source would be far worse than the bug"* in the PR body and then did precisely that.

The tests written alongside that commit did not catch it because they asserted only the extensions the fix happened to allow. A test that enumerates the implementation's own answer cannot detect the implementation being wrong about the world — the same shape as the `github-actions[bot]` fixture bug in #266.

## Root cause of both bugs

The predicate existed **twice** — `packages/hollow-test/lib/targets.mjs:10` and `scripts/mutation-gate.mjs` — carrying the identical exclusion regex, and they drifted. Drift here is silent in the dangerous direction: if the wrapper decides a file is not source, hollow-test is never invoked for it at all.

Fixing only the wrapper would have left two copies to drift again. So there is now **one exported predicate**, `isMutableSource`, defined in `targets.mjs` and imported by the wrapper, include-listing the JS and TS families.

## The load-bearing test

Not the extension lists — the **cross-contract test**:

```js
test("the wrapper and hollow-test agree on what source is", () => {
  for (const f of cases) {
    assert.equal(hollowTestWouldMutate(f), isMutableSource(f), ...);
  }
});
```

It asserts the two selectors agree across JS, TS, extensionless, dotfile, non-code and test/spec paths, rather than trusting two lists to be maintained in step. Plus a `classify()`-level case proving a TypeScript-only diff does not short-circuit to `nothing`.

## Mutation-verified in both directions

```
drop the TS family (the review finding):
  not ok 4 - hollowTestWouldMutate admits every source extension hollow-test mutates
  not ok 6 - a TypeScript-only diff still reaches the mutation gate

revert to the exclusion list (the original bug):
  not ok 2 - hollowTestWouldMutate does not treat extensionless files as source
  not ok 3 - hollowTestWouldMutate does not treat dotfiles as source
```

75 tests pass across the wrapper and hollow-test suites.

## Note on #288

Restoring TS to the source set means a docs-only `.tsx` change can again reach the slow path, where the baseline is broken. That is not a regression from this PR — it was the behaviour before the first commit — but it does mean #288 gates more than it looked like it did.

voodootikigod:
## Final state — consolidating 14 commits and 10 review rounds

CI: **all 20 checks green**, `mutation-gate` included (it now validates its own fix, 12/12 mutants killed). Local: 94 tests across both suites, full suite 46/46.

The title was renamed — this started as a one-line filter fix and ended up rewriting hollow-test's target selection. Worth knowing before reading the diff.

### What actually changed

| | Before | After |
|---|---|---|
| Source definition | Two copies (wrapper + `targets.mjs`) that drifted | One exported `isMutableSource` |
| Classification | Exclude guessed non-source extensions | Allow-list: `.mjs`, `.cjs`, `.js` |
| Test detection | `/(test\|spec)/` substring over the whole path | Path segments + `node --test` basename conventions |
| Ambiguous names | Guessed | Declared via `--source-glob` / `--test-glob` |
| Explicit targets | Bypassed every check | Bypass test-path rules only; language guard enforced |
| Skipped files | Silent | Reported on every classification |

### The bugs, in the order they surfaced

1. **Extensionless files classified as source** — `CODEOWNERS` alone stranded the gate on the full-suite slow path. `LICENSE`, `Dockerfile`, `.gitignore` too.
2. **My first fix dropped 49 `.ts`/`.tsx` files** while hollow-test still accepted them → `classify()` returned `nothing` and the required gate exited 0 without running. A false negative, strictly worse than the bug.
3. **Substring matching hid 11 production files** — the entire `hollow-test` package (the mutation tool exempting itself), all of `spec-lint`, and `gate-manifest/lib/attest.mjs`.
4. **`--target`/`--rails` bypassed the language guard**, then my guard over-corrected and broke the P3 rails workflow — rails *are* test files.
5. **Post-run verifier checked the unfiltered set**, so a mixed rails expansion did the work then failed.
6. **`.py` claimed but unimplemented** — zero Python operators, and `test_*.py` would be mutated as production source, self-killing. Declined; #292.
7. **Hyphenated names are `node --test` conventions**, not stylistic. My round-5 reasoning was wrong.
8. **Invalid mutants credited as kills** — the root cause; #293.

### Two hollow tests I wrote, both caught by mutating my own guards away

- The unsupported-`--target` test asserted only "non-zero exit and mentions the file" — which passes *with the guard removed*, because hollow-test then mutates the `.py`, mutants survive, and it exits 2 naming the file.
- The `--source-glob` test passed `--target`, bypassing the very classification path it claimed to verify. Dropping `sourceGlobs` from `filterTargetFiles` left it green.

Both now assert the specific behaviour and fail when the guard is removed. Every fix in this PR is mutation-verified in both directions.

### Not fixed here, deliberately

- **#293** — invalid mutants scored as kills. Round 10 showed this is *not* TS-specific: `null-return` on a multiline `return {` yields invalid JavaScript, the parse failure exits non-zero, and it is credited as a kill. At quota one it can be the only trial. The allow-list narrowed the trigger; the scoring bug is language-independent, and `review-calibration` shares the assumption. **This is the root cause and it outranks everything in this PR.**
- **TypeScript-only diffs exit 0 with a printed skip.** Review argued they should fail. That is a policy choice with real blast radius — every `apps/docs` PR would fail a required check until #293 lands — so I left it green-and-reported rather than deciding it unilaterally.

### Breaking change

`@adlc/hollow-test` is published, and `filterTargetFiles` narrowed: non-JS source is no longer mutated. Deliberate, on the principle that declining support beats claiming support the operators do not implement. Both doc copies state the real contract.

Issue #292, opened 2026-07-22
hollow-test: add real Python support (operators + pytest test discovery)


voodootikigod:
Split out from PR #289, where I briefly added `.py` to the source allow-list and then backed it out.

## The claim

`packages/core/lib/mutate.mjs` header:

```
// Text-based by design: language-agnostic-enough for JS/TS/Python-style code,
```

Both `packages/hollow-test/README.md` and `docs/tools/hollow-test.md` previously described the filter as an exclusion list, which implicitly admitted `.py`.

## Why the claim does not hold

**1. There are no Python-aware operators.** Every operator in `mutate.mjs` is JS-shaped — `invert-comparison` (`===`/`!==`), `bool-flip` (`true`/`false`), `null-return` (`return null`), `logic-swap` (`&&`/`||`), `negate-guard-subclause`, `off-by-one`. Nothing handles `return None`, `True`/`False`, `and`/`or`, or `elif`. `grep -c "None\|def \|elif" packages/core/lib/mutate.mjs` → `0`.

**2. Python test discovery is missing, and this one is dangerous.** The filename exclusion recognizes the dotted JS convention (`*.test.*`, `*.spec.*`) but not Python's, where `test_*.py`, `*_test.py`, and `conftest.py` **are** the tests. Admitting `.py` therefore classifies pytest modules as production source. Mutating an assertion inside a test makes that test fail, which scores the mutant **killed** — a vacuously green gate that proves nothing.

I nearly shipped exactly that: an intermediate commit on #289 admitted `.py` and pinned `src/test_app.py` as mutable source. Caught in adversarial review.

## Current state after #289

`.py` is out, and `SOURCE_EXT_RE` is JS/TS only. Both documentation copies now state the real contract. This narrows published `filterTargetFiles` behaviour for any consumer mutating non-JS source — a deliberate, documented breaking change, on the reasoning that **declining support is better than claiming support the operators do not implement**.

## What real Python support would need

1. Python mutation operators: `return None`, `True`/`False`, `and`/`or`, comparison inversion, `elif` guards.
2. Python test discovery in the exclusion: `test_*.py`, `*_test.py`, `conftest.py`, and `tests/` segments.
3. An end-to-end Python fixture proving both directions — a production mutant survives without a test, and a test-only diff does not self-kill.
4. A decision on whether `filterTargetFiles` should take a configurable pattern set rather than hard-coding conventions per language.

[...]

PR #294, opened 2026-07-22
feat(adlc-antigravity): native integration parity build-out (#241)

voodootikigod:
## Summary

Establishes feature and enforcement parity between the native Antigravity plugin (`plugins/adlc-antigravity/`) and `antigravity-booster`, fulfilling the approved parity plan for [#241](https://github.com/voodootikigod/adlc/issues/241).

### Key Features & Components Built:

1. **P5 Prosecutor Roster (`plugins/adlc-antigravity/agents/`)**:
   - 5 specialized prosecutor lens subagents (`prosecutor-contract.md`, `prosecutor-correctness.md`, `prosecutor-diff.md`, `prosecutor-security.md`, `prosecutor-tests.md`) and 1 verifier agent (`prosecutor-verifier.md`).

2. **In-Session Build-Gate Backstop (`build-gate-inline.mjs`)**:
   - Persistent depth tracking across process executions (`.adlc/sessions.json`) with atomic temp-file writes and PID-probed lock reclamation.
   - Fail-closed evaluation (`deny`) for high-risk ticket builds when context depth or compaction thresholds are crossed under `ADLC_P4_ENFORCEMENT=1`.
   - Fail-closed security protection against ticket pointer conflicts and corrupted ticket stores.

3. **In-Session Flail Detection (`flail-inline.mjs`)**:
   - Zero-dependency file edit churn warning emitted to stderr when a single file is rewritten 3+ times in a session.

4. **Diagnostic Commands & Packaging**:
   - `/adlc-status` and `/adlc-doctor` CLI entrypoints with ancestor ADLC root resolution.
   - Packaging allowlist and test script updates (`package.json`, `packaging.test.mjs`).

5. **Self-Orchestration Skill Guidance (`skills/adlc-self-orchestrate/SKILL.md`)**:
   - Updated model pool hint routing (`gemini` | `claude` | `auto`).

---

### Quality & Verification:
- **Unit Test Suite**: **88 / 88 tests passing green** (`npm test`).
- **Adversarial Review Loop**: 14 passes executed until `APPROVE` verdict achieved with zero open findings.

PR #290, opened 2026-07-22
feat(copilot): build plugins/adlc-copilot, the 7th ADLC harness (#242)

voodootikigod:
## Summary

Implements #242 — `plugins/adlc-copilot`, the 7th native ADLC harness — built to the contract **verified against the real GitHub Copilot CLI 1.0.73** (per the #240 probe), not the integration plan's pre-probe assumptions.

The probe overturned three plan claims (see `docs/integrations/copilot-probe-appendix.md`):
- **Deny shape:** a `preToolUse` hook denies by printing a non-empty `{"reason":…}` object to stdout + exit 0 — NOT `permissionDecision`, NOT exit 2.
- **Hooks fail OPEN** on a crashed/timed-out hook (`decision=undefined` → tool proceeds). The plan's "fail-closed, stronger than agy/Cursor" is false.
- Entry shape is `bash`/`timeoutSec`; stdin is camelCase `{toolName, toolArgs(JSON string), cwd}`.

## What shipped
- **Plugin** `plugins/adlc-copilot/`: manifest, hooks.json, rails-guard + build-gate + lifecycle adapters (deny via `{reason}`, never crash → application-level fail-safe, deny written synchronously to fd 1 so it can't be truncated into an allow), generated `@adlc/core` readers, 6 skills, 6 read-only prosecutor agents, `.mcp.json` + shims.
- **Marketplace** `.github/plugin/marketplace.json` (`copilot plugin marketplace add voodootikigod/adlc`).
- **Install smoke** `scripts/copilot-install-smoke.mjs` — offline contract + behavioral deny-shape proof, wired into `ci.yml` on every test leg; live leg gated on `ADLC_COPILOT_LIVE_INSTALL=1`.
- **Fleet adapter** `packages/fleet/lib/adapters/copilot.mjs` + registration + tests.
- **`adlc init --harness copilot`** scaffolding.
- **Docs:** `copilot.md`, ADR-0013, `harness-capability-matrix.md`, `package-reference.md`.

## Honest posture
The in-session hook is **advisory-tier** (Copilot fails open on process death/timeout); the `rails-guard-ci` diff gate is the **unbypassable enforcement tier**. The live deny-proof (#240 live leg) could not run — an org Copilot policy blocks model turns — and must run in an unrestricted account/CI.

**AC1 needs rewording:** its "hook internal error → deny (fail-closed)" pin contradicts the verified binary; see the AC1 correction comment on #242 and ADR-0013.

## Test plan
[...]

PR #306, opened 2026-07-22
fix(hollow-test): a kill must mean the tests ran and failed (#293)

### LINKED ISSUE #293: hollow-test: invalid mutants are credited as kills, blocking TypeScript/JSX support
Split out from PR #289, which excluded `.ts`/`.mts`/`.cts`/`.tsx`/`.jsx` (and earlier `.py`) from the mutation allow-list because of this.

## The defect

`packages/hollow-test/lib/runner.mjs`:

```js
const killed = trial.timedOut || (trial.status !== 0);
```

**Any** non-zero exit counts as a kill — including a parse or compile failure. So a mutant that produces syntactically invalid code is scored as successfully killed, when in fact nothing was tested.

## Why it blocks TypeScript and JSX

The operators in `packages/core/lib/mutate.mjs` are text-based regexes. `invert-comparison` cannot distinguish a comparison from a type argument or a JSX delimiter:

```
Promise<unknown>   →   Promise>=unknown>      (invalid)
<Button>           →   >=Button>              (invalid)
```

With the common per-file quota of one, that single invalid mutant consumes the entire budget: no survivor remains, and the required gate exits 0 having exercised no behavioural assertion. The gate built to detect false greens produces one.

This is not TS-specific in principle — it is why `.py` was also declined (#292) — but TS/JSX is where it bites in practice, because generics and JSX are common and `<`/`>` are everywhere.

## Current state after #289

The allow-list is plain JavaScript only: `.mjs`, `.cjs`, `.js`. The exclusion is **reported** by `scripts/mutation-gate.mjs` rather than silent:

```
mutation-gate: 1 changed file(s) are not a mutable source type, so this gate does not cover them:
  plugins/adlc-pi/index.ts
```

That is a documented coverage boundary rather than a hole — but it *is* a boundary, and this repo has 49 tracked non-test `.ts`/`.tsx` files (mostly `apps/docs`) that the gate does not cover.

## What a fix needs

1. **Validate the mutant before scoring it.** A mutated file that does not parse must be *discarded*, not counted as killed. This is the root fix and it also removes the language coupling — an invalid mutant is meaningless regardless of language.
2. **Syntax-aware operators**, or at minimum a parser-backed guard so `invert-comparison` never fires inside a type argument or JSX element.
3. Then re-admit `.ts`/`.mts`/`.cts`/`.tsx`/`.jsx`, with regression tests at quota one proving `Promise<unknown>` and JSX markup cannot produce a passing gate.

`acorn` is already a devDependency in this repo, though it does not parse TS/JSX without plugins — worth deciding whether the validation step belongs in `@adlc/core` (shared with review-calibration, which has the same scoring assumption) and what parser it should use.

## Related

- #292 — Python declined for the same reason plus missing test discovery.
- The scoring assumption is shared with `packages/review-calibration`, which should be checked for the same false-kill path.

voodootikigod:
Ticket T62. Fixes #293, and unblocks #292.

## The defect

`runner.mjs` inferred a kill from a non-zero exit — and a file that does not **parse** also exits non-zero. So an unparseable mutant was recorded as a successful kill while nothing was tested.

Reproduced on **ordinary JavaScript**, not TypeScript. `null-return` rewrites a multiline `return {` to `return null;` and strands the object literal's remaining lines:

```
before:  Total: 3  Killed: 3  Survived: 0                 exit 0
after:   Total: 3  Killed: 2  Survived: 0  Invalid: 1
```

At the common per-file quota of one, that invalid mutant is the **only** trial — so the required gate exited 0 having exercised no assertion at all. A false green from the tool built to detect false greens. The JS-only allow-list in #289 removed the most frequent trigger (`Promise<unknown>` → `Promise>=unknown>`); this removes the cause.

## The principle

**A kill must mean the tests ran and failed.** Every finding in this PR is a place where something *other* than a failing test was being counted as evidence. Review found four such layers, each below the last:

| Layer | What was credited as a kill |
|---|---|
| Mutant | Code that did not parse |
| Syntax checker | A checker that could not run (assumed "valid") |
| Test runner | A spawn failure (`EAGAIN`/`ENOMEM`), labelled "timed out" |
| Shell | `/bin/sh` starting but failing to exec the test binary (126/127) |

All four now resolve to **undetermined** — never a kill, never a survivor — and carry a reason. Undetermined is an operational failure, because guessing is what reopened the bug each time.

Safe to treat 126/127 as a launch failure because the **green baseline already proved this exact command can launch**; a suite genuinely exiting 127 would have failed the baseline first, loudly. Ordinary non-zero exits (1, 2, 125, 128) remain real test failures, asserted so this cannot quietly widen.

## Accounting

- Invalid mutants are **discarded** — neither killed nor survived. Counting one as a kill fakes coverage; counting it as a survivor blames the tests for code that was never valid.
- The all-invalid guard is **per file**, not global. A global check passes the moment any *other* file yields a kill, which would let a named `--target` go entirely untested.
- Accounting is seeded from the **selected targets**, not from results, so a file that produced nothing cannot slip through as "no news is good news".
- A file with zero quota is **reported**, not fatal — reserving budget for explicit targets is the documented design (#70/#41/#35). Refusing outright broke an existing test; silence would make an unprosecuted file the ideal place to hide a change.

## Verification

- `node --check` rather than a parser dependency: `@adlc/hollow-test` is published, and this asks exactly the right question (would Node accept this file), honouring the real extension and package type.
- `classifyTestResult` is split out and exported so **every** branch is testable with a synthetic result. `EAGAIN`/`ENOMEM` cannot be provoked reliably in a test — and "cannot be provoked" is precisely how this stayed unnoticed — so the tests supply the shapes directly.
- `mutation-gate`: **12/12 mutants killed**. 117 tests across the affected suites. Full suite 46/46.
- Every fix is mutation-verified: removing the syntax gate, treating `unknown` as valid, folding spawn failures into `timedOut`, and narrowing the per-file guard each fail a specific test.

## Course corrections during review

Worth flagging, since both were mine:

- I first refused any run whose budget could not cover every target file. That broke the deliberate reserve-for-explicit-targets design (#70/#41/#35) and an existing test caught it. The two zero-valid cases are genuinely different and are now handled separately.
- My first warning said a file "received no mutation budget" when it had budget and simply had no mutable lines (all comments). False alarms are how real warnings stop being read.

## Breaking-ish

The JSON report gains `invalid` / `undetermined` buckets and a `reason` field, and `checkFailed` was renamed `undetermined` before release. Both doc copies updated.

## Test plan

- [x] `node --test packages/hollow-test/test/*.test.mjs` — all pass
- [x] `node scripts/mutation-gate.mjs --base main` — 12/12 killed
- [x] `npm test` — 46/46 segments
- [x] Live reproduction confirms the invalid mutant is no longer a kill
- [ ] After merge: consider re-admitting TS/JSX (#293 was the blocker) and Python (#292)


voodootikigod:
@copilot please fix the merge conflicts in this pull request.

voodootikigod:
Merged current `main` in and resolved what it brought with it.

## Answering the conflict question

**#287 needed nothing** — it is already merged (`3d4c845` on main).

**#306 had no textual conflict with the main you asked about.** Confirmed three ways: `git merge-tree` reported a clean merge, there was zero file overlap between the two sides, and the merge itself resolved cleanly.

But `main` moved twice more while I was working (#290 copilot, then #304 fleet and #294 antigravity), and the second wave *did* conflict. Three things came out of it:

### 1. A ticket ID collision

**#304 landed a different `T62`** — `fleet: make checkFlail actually read flail-detector's real contract` — while this branch was open. Two unrelated tickets claiming one id.

Main's T62 is kept; this work is renumbered **T63**. The `.adlc/tickets.json` conflict was resolved to main's version and T63 re-created through `adlc ticket create`, so `planHash`/`storeHash` are computed from the contents rather than hand-merged — those hashes are what the gates trust.

### 2. A local branch that had drifted behind origin

My local copy of this branch was based on an *older* main than origin's: same 8 commits, different SHAs, and origin already carried #297. The nine files were byte-identical between them, so I reset local to origin rather than force-pushing a stale rewrite over the version CI had already validated.

### 3. Why the merge was worth doing at all

Running the gate against the newer main surfaced survivors in `packages/fleet/lib/gates.mjs` and `live-deps.mjs` — files this PR does not touch. They came from #304, and they were only visible because the gate was diffing against a main I had not merged yet. Not this PR's problem, but worth a look on #304: `array-literal-shrink` dropping `--json` from the flail-detector args, and `timedOut: false` → `true` in a catch block, both survived.

## State after the merge

```
node --test packages/hollow-test + mutation-gate    119 pass, 0 fail
mutation-gate --base origin/main                    12/12 killed
```

The PR body still says "Ticket T62" — read that as **T63**.

Issue #284, opened 2026-07-22
fleet: checkFlail reads a JSON field (`detected`) flail-detector never emits — flail consultation always fails open


voodootikigod:
## Problem

`packages/fleet/lib/gates.mjs:70` reads `parsed.detected === true` from flail-detector's `--json` output to decide whether a build session is flailing between strikes:

```js
const out = exec(adlcBin, args); // adlc flail-detector <log> --json
const parsed = JSON.parse(out);
return { flail: parsed.detected === true, signals: parsed.signals ?? [] };
```

But flail-detector's real JSON output (`packages/flail-detector/lib/analyze.mjs` → `bin/flail-detector.mjs`'s `printJson(result)`) has never had a `detected` field. It has always been `{ verdict: 'flail' | 'clean', signals: [...], bytes, recommendation? }`.

Confirmed by direct repro:

```bash
$ echo 'Writing /etc/passwd' > log.txt
$ node packages/flail-detector/bin/flail-detector.mjs log.txt --scope 'src/**' --json
{
  "verdict": "flail",
  "signals": [{ "type": "scope-violation", "entries": [{ "path": "/etc/passwd", "line": "Writing /etc/passwd" }] }],
  "bytes": 20,
  "recommendation": "Kill the session. Review signals above and regenerate fresh."
}
```

`parsed.detected` is `undefined` here, so `checkFlail` returns `{ flail: false }` even though the tool clearly detected a flail signal (`verdict: "flail"`). This means **fleet's flail-detector consultation between build strikes is a silent no-op in production** — it always fails open (§12's documented fallback for detector *errors*, not for a detector that ran fine and returned a positive verdict under a mismatched field name).

`fleet`'s own tests never catch this because they mock the exec output directly with the wrong shape too — `{ detected: true, signals: [...] }` (`packages/fleet/test/gates.test.mjs:40`, `packages/fleet/test/live-deps.test.mjs:22`, `packages/fleet/test/live-concurrency.test.mjs:31`) — so the tests are internally consistent with the bug, not with the real tool.

## Impact

[...]

PR #316, opened 2026-07-23
fix(fleet): make a fail-open flail consultation observable (#309)

### LINKED ISSUE #309: fleet: a fail-open flail consultation is indistinguishable from a clean verdict — the gate can go blind silently
## Problem

`checkFlail` returns `{ flail, signals, failedOpen? }`. The `failedOpen` flag marks the §12 backstop — the detector could not produce a trustworthy verdict, so the gate deliberately declines to cut the build short.

**Nothing reads it.** `scheduler.mjs` consults `.flail` alone:

```js
// packages/fleet/lib/scheduler.mjs:74 and :84
if (canRetry() && (await effects.flail({ ticket })).flail) {
  return fail('flail-detector diagnosed a genuine flail — skipping the second strike');
}
```

A repo-wide search for `failedOpen` finds its construction in `gates.mjs` and assertions in `packages/fleet/test/*` — no production consumer.

So at runtime, **every** fail-open is indistinguishable from a genuine clean verdict:

- the detector is not installed, or is unspawnable
- an operational error (exit 1)
- output that will not parse
- a document whose `verdict` we no longer recognize (schema drift)
- `statusDir` is unwritable, so no transcript exists — `appendLog` is best-effort by design and swallows the failure

Each of these silently reports "this session is fine, keep building."

## Why this matters

This is the same failure class as #284, which is the whole reason that issue was expensive: a supervision control stopped working and kept reporting normally. #284 fixed *why* the gate was blind. This is about *noticing* when it goes blind again.

The fail-open policy itself is correct and should not change — an unverifiable signal must not cut a build's normal retry short, and the two-strike cap remains the backstop. The problem is purely that it is unobservable. An operator running a fleet today has no way to learn that flail-detector has been silently disabled for every ticket.

Concretely: mistype `statusDir`, or run in an environment where `@adlc/flail-detector` fails to resolve, and every ticket in every run gets a clean flail verdict forever, with no log line, no status field, and no gate-manifest entry.

## Proposed fix

Make the fail-open observable without changing the policy:

- Have `checkFlail` include a machine-readable `reason` on the fail-open (`spawn-error` / `operational-error` / `unparseable` / `unrecognized-verdict`), so the cause is not flattened away.
- Have the scheduler `log()` a warning when `failedOpen` is true, naming the ticket and the reason, rather than silently treating it as clean.
- Consider recording it to the gate manifest / fleet status so a run's evidence shows the gate was consulted but could not answer — a consultation that could not be performed is not the same as one that passed.

## Acceptance criteria

- [ ] `checkFlail`'s fail-open carries a `reason` distinguishing the causes listed above, covered by unit tests per branch
- [ ] A fail-open during a run emits an operator-visible warning naming the ticket and reason; verified by a test asserting the emitted log, not just the returned object
- [ ] A genuine `clean` verdict emits no such warning (no crying wolf on the happy path)
- [ ] The fail-open POLICY is unchanged: `flail` stays `false` and the build still gets its normal retry — assert this explicitly so the fix cannot drift into fail-closed
- [ ] `node --test packages/fleet` green

## Notes

Found during P5 prosecution of #284 (PR #304) — three independent lenses flagged it. Filed separately rather than widening that PR a third time; #304 exists because a bug survived weak review, and growing it works against its own purpose.

Scope is `packages/fleet/lib/gates.mjs` + `packages/fleet/lib/scheduler.mjs` (+ tests). Depends on #304 for the `unrecognized-verdict` fail-open path.

voodootikigod:
Fixes #309.

Follow-on to #284 / #304 (now merged). Rebased onto current `main`.

## Problem

`checkFlail` returned `{ flail, signals, failedOpen? }`, and **nothing read `failedOpen`**. `scheduler.mjs` consulted `.flail` alone. So at runtime every fail-open was indistinguishable from a genuine clean verdict:

- the detector is not installed, or is unspawnable
- an operational error (exit 1)
- output that will not parse
- schema drift — a `verdict` we no longer recognize
- `statusDir` is unwritable, so no transcript exists (`appendLog` is best-effort by design and swallows the failure)

Each silently reported "this session is fine, keep building." Mistype `statusDir`, or run somewhere `@adlc/flail-detector` fails to resolve, and every ticket in every run gets a clean flail verdict forever — no log line, no evidence entry, no way for an operator to find out.

This is the same failure class as #284, which is exactly why that bug was expensive: a supervision control stopped working and kept reporting normally. #284 fixed *why* the gate was blind. This fixes *noticing* when it goes blind again.

## The policy is unchanged

Worth stating plainly, because the tempting "fix" here is the wrong one: **this does not make the gate fail closed.** An unverifiable signal still must not cut a build's normal retry short, and the two-strike cap remains the backstop. `flail` stays `false` and the build keeps its retry. Only the *visibility* changes.

Failing closed would mean a missing detector starts killing real work — a worse failure than the one being fixed. There is an explicit test pinning this, and the mutation run confirms flipping it to fail-closed reds that test.

## Changes

- `failOpen()` carries a machine-readable `reason`: `spawn-error`, `operational-error`, `unparseable`, `unrecognized-verdict`. A numeric exit status means the detector **ran** and failed operationally; anything else means we never got a usable process at all — collapsing those two would hide a missing install behind a plausible-looking error.
- Both scheduler consultation sites route through `consultFlail()`, which warns via the injected `log` (naming ticket and reason) when the gate could not answer, and stays quiet on a real verdict.
- The effects-contract JSDoc documents the real return shape (`flail({ticket}) → { flail, signals?, failedOpen?, reason? }`) instead of `{ flail }`.

## Verification

- `node --test packages/fleet/test/*.test.mjs` — **257/257 pass**
- `adlc hollow-test --base main` — **16/16 mutants killed**
- `adlc rails-guard --ticket T63` — clean
- Full repo suite: **51/51 segments, all green**

Both drift directions are pinned by test, verified by re-introducing each:

| Mutation | Result |
|---|---|
| Delete the warning (back to silent) | reds the observability test |
| `flail \|\| failedOpen` (fail-closed) | reds the policy test |

Mutation prosecution also surfaced a **pre-existing coverage gap**: the scheduler consults flail at two sites (failed build, failed gate), and every existing test failed at `dispatch`, so the gate-failure site was never reached. That left its `canRetry() && consultFlail()` mutable to `||` — "always diagnose a flail while retries remain", killing tickets that never flailed — with the entire suite green. Now covered.

Two `deepEqual` assertions introduced by #304 were **tightened** to assert the new `reason` rather than loosened to ignore it.

## ADLC

Ticket **T64**, P0 → P5. Scope: `packages/fleet/lib/gates.mjs`, `packages/fleet/lib/scheduler.mjs`, `packages/fleet/test/**`. flail-detector's contract is the frozen rail and is untouched.


voodootikigod:
@copilot resolve the merge conflicts in this pull request

Issue #326, opened 2026-07-23
Trust-root cross-model review is a built, tested, fail-closed gate that no CI job ever invokes


voodootikigod:
## Problem

T39 / ADR-0007 made cross-model adversarial review a **gate** for the trust-root tier rather than a recommendation. The classifier exists (`packages/prosecute/lib/tier.mjs`), fails closed, is tested, and the P5 runner exits `2` until the manifest holds a matching `cross-model-review` approve.

**Nothing invokes it on a pull request.** The gate is fully built and never runs.

## Evidence

**1. No CI job runs the P5 gate.** The only `prosecute` references in `.github/workflows/` are harness live-proofs — `scripts/opencode-live-prosecute.mjs` (T33) and `scripts/pi-live-prosecute.mjs` (T26 AC5). Both exercise a harness's tool wiring; neither prosecutes the PR's own diff.

The 20 checks that run on a PR:

```
codex-live-latest  mutation-gate  opencode-live-latest  pre-ga-gate  rails-guard
router-drift  test (18|20|22)  ticket-store-platform (3 os × 3 node)  Vercel ×2
```

No `prosecute`, no `run p5`, no cross-model check.

**2. The tier is far broader than it looks.** `TRUST_ROOT_FILES` includes `.adlc/tickets.json`, and that check is **unconditional** — the test-file exemption applies only to the package-prefix surfaces:

```js
if (TRUST_ROOT_FILES.includes(path)) {
  push(`touches trust-root file ${path}`);
}
```

Adding a ticket modifies `.adlc/tickets.json`. So **every ADLC-ticketed PR is trust-root tier** and requires a cross-model attestation by the current rule.

**3. Observed consequence.** Four PRs merged this week — #304, #316, #317, #323 — each adding a ticket, and #317 additionally changing `packages/rails-guard/bin/` (an `ENFORCEMENT_PREFIXES` surface, non-test). All four were trust-root tier. None has a recorded `cross-model-review` attestation. All merged green, because no check ever asked.

## Why this is the same failure class as #284, #309 and #322

A mechanism that reports success while enforcing nothing.

CI is green, the manifest records that gates ran, and the trust-root requirement is silently absent. As with #322, the absence is invisible: there is no signal distinguishing "cross-model review passed" from "cross-model review was never requested." The gate's own fail-closed design is irrelevant if the fail-closed path is never entered.

## The tension worth resolving

These two facts are hard to hold together:

[...]

PR #354, opened 2026-07-25
harden(#326): sign-verified cross-model gate, functional on a mixed manifest, key isolated from PR code

voodootikigod:
Follow-up hardening on the trust-root cross-model review gate that shipped in #349. As merged, the gate was **forgeable**, and — once a signing key was introduced — it would have been **inert** (and un-recordable) on this repo's real manifest. This closes both, and isolates the signing key from PR-controlled code.

## Why

#349's gate matched an attestation's self-reported fields. Since the manifest lives in the PR-controlled tree, a contributor could append their own `approve` and clear the gate (found by adversarial-review, high/0.99). The natural fix — HMAC-sign attestations and verify — ran into two non-obvious couplings in `gate-manifest`:

1. **`verify()` (read) requires *every* entry signed when a key is present.** The shared `manifest.jsonl` has **88 historical entries, all unsigned** (signing was never enabled repo-wide). So a keyed `verify()` fails on entry #1 → the gate would fail closed for **every** PR, valid attestation or not. An always-failing gate just gets bypassed — it enforces nothing.
2. **`record()`'s pre-append check has the same requirement** → the maintainer couldn't even *record* an attestation onto the real manifest (it throws "unsigned entry").

## What

**Forge resistance** — `candidateReview()` verifies each attestation's HMAC signature (`verifyEntrySig` under `ADLC_MANIFEST_KEY`) before trusting it. The `tier-check` CLI requires the key and **fails closed** (exit 2, distinct message + `keyAvailable:false`) without it.

**Functional on a mixed manifest** — `verify()` and `record()`'s pre-append check gain an additive `requireSignatures` option (**default `true` — no change for existing callers**). The cross-model gate uses chain-integrity-**only**: it still catches the corrupt-and-skip revocation attack (#349 Codex F2 — a garbled line breaks the hash linkage) while tolerating unsigned history. Provenance is enforced **per-entry** on just the approve the gate is about to trust. Chain-only + per-entry-sig = tamper-evident **and** usable.

**Key isolation** — the folded step ran the PR's own checked-out code with the secret. It's removed; the gate moves to a separate `pull_request_target` workflow (`cross-model-gate.yml`) that runs **BASE** code + BASE deps against the PR tree as **DATA** (git worktree, never executed). The key is seen only by trusted base code. Fork PRs get no secret ⇒ fail closed (a fork touching trust-root code needs a maintainer).

## ⚠️ Required setup (or the gate fails closed for every trust-root PR)

- Set the **`ADLC_MANIFEST_KEY`** GitHub Actions secret.
- Record attestations locally with that key set so the entry is signed.

## Test plan

- [x] `packages/prosecute` — 128/128 (adds: signature-verification unsigned/wrong-key/no-key; unsigned-legacy-manifest **not inert** + forge still rejected; per-clause fail-closed input guard; forge + `--json` no-key CLI contract).
- [x] `packages/gate-manifest` — 69/69 (additive `requireSignatures`, defaults unchanged).
- [x] Full suite — 1705/1705, 0 fail.
- [x] `mutation-gate` — 16/16 mutants killed on the changed reader/CLI.
- [x] `cross-model-gate.yml` + `ci.yml` — valid YAML; no test pins the removed folded step; install-hardening (`--ignore-scripts`) contract intact.
- [ ] Cross-model adversarial review (codex, distinct provider) — running.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -29,15 +29,17 @@ jobs:
 
       - name: Install
         # #326 Codex F1 (round 3): install with NO lifecycle scripts, then rebuild ONLY
-        # a trusted allowlist of native deps. This is the job that runs the trust-root
-        # cross-model gate below; a candidate `postinstall` in ANY workspace/dep manifest
-        # would otherwise run here and could shim node/npm onto $GITHUB_PATH (persisted
-        # to later steps), defeating the gate. The repo has NO first-party lifecycle
-        # scripts, and the ONLY deps that declare one are these lockfile-pinned native
-        # packages (verified via package-lock `hasInstallScript`), so this reproduces a
-        # normal install for any legitimate PR while never running candidate-controlled
-        # code. A candidate cannot extend the allowlist without a lockfile change, which
-        # tiers. Keep this list in sync with `hasInstallScript` deps.
+        # a trusted allowlist of native deps. This job checks out the PR and runs its own
+        # test suite (`npm test`), i.e. candidate-controlled code; a candidate `postinstall`
+        # in ANY workspace/dep manifest would otherwise run here and could shim node/npm
+        # onto $GITHUB_PATH (persisted to later steps) or tamper with the runner. The repo
+        # has NO first-party lifecycle scripts, and the ONLY deps that declare one are these
+        # lockfile-pinned native packages (verified via package-lock `hasInstallScript`), so
+        # this reproduces a normal install for any legitimate PR while never running
+        # candidate-controlled install code. A candidate cannot extend the allowlist without
+        # a lockfile change, which tiers. Keep this list in sync with `hasInstallScript` deps.
+        # (The trust-root cross-model gate no longer runs in this job — it moved to the
+        # trusted-context cross-model-gate.yml so it never touches PR-controlled code.)
         run: |
           npm install --ignore-scripts
           npm rebuild esbuild sharp msgpackr-extract protobufjs @google/genai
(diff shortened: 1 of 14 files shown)


voodootikigod:
## Cross-model review round (codex, gpt-5) — addressed

The distinct-provider adversarial review returned two findings; here's how each is handled.

**F1 (high) — rewrite/truncate a signed revocation to resurrect an approval.** Two halves:
- **Rewrite (closed in code):** my initial `requireSignatures:false` was too permissive — it skipped signature validation even for entries whose sig is *present but invalid*, so an attacker could rewrite a signed `needs-attention`, invalidate its sig, recompute public prev-hashes, and pass chain-only verify. Fixed: `verify()`/`record()` lenient mode now **tolerates missing (legacy) sigs but still rejects any present-but-invalid sig** — "unsigned history is OK, tampered signatures are not." New tests: `gate-manifest` lenient-mode tamper rejection; `cross-model` end-to-end rewrite-a-signed-revocation stays failed-closed.
- **Truncation (documented honest-limit):** a hash chain has no authenticated head, so a branch-owner can *drop* a signed revocation recorded on their branch. This is fundamental to a PR-controlled append-only log; base-anchoring can't protect post-divergence attestations. It's closed operationally by **F2** (required check under branch protection) or an external signed checkpoint. Documented at `manifestChainTrustworthy`.

**F2 (medium) — gate is advisory, not enforced.** True and now reframed: `cross-model-gate.yml` is designed to be marked a **required status check** via branch protection on `main` (it fails closed, so as a required check it blocks the merge — which also closes the truncation limit). The separate-workflow design is *compatible* with being required; the folded-into-required-job wiring was what forced the unsafe key exposure. Until branch protection is configured this repo has none, so the gate is an intentional fallback signal.

### ⚠️ Deploy steps to make this binding
1. Set the **`ADLC_MANIFEST_KEY`** Actions secret; record attestations with it.
2. Mark the `cross-model-gate / gate` job a **required status check** (branch protection / ruleset on `main`).

Verification after this round: gate-manifest 70/70, prosecute 129/129, mutation-gate 20/20 killed (deep pass). Full suite re-running.

voodootikigod:
## Round 2 review outcome (codex) — rewrite closed; truncation over-claim corrected

The distinct-provider review **confirmed the rewrite fix is correct** ("correctly rejects a present-but-invalid signature while tolerating missing legacy signatures, so rewriting a signed revocation is closed"). The deterministic gate derived `approve` (exit 0), but the model raised a fair point I've now corrected:

- **My comments overstated the control.** I claimed a required check under branch protection *closes* truncation. It does **not** — the required check evaluates the PR-controlled (truncated) tree, finds a valid chain ending in the earlier signed approve, and passes. I've corrected `manifestChainTrustworthy()` and `cross-model-gate.yml` to say so plainly.
- **Truncation is genuinely out of code-scope here.** Because the dropped revocation is a *PR-branch* entry, base-anchoring can't see it. Closing it requires the latest attestation state anchored **outside** the PR-controlled tree (a protected ref / check-run / external signed checkpoint) — i.e. moving where attestations are stored. That's a design change, tracked as follow-up, not a patch.

**Net for this PR:** forgery (unsigned/wrong-key) and **rewrite** (tampered sig) are closed in code + tests; **truncation** and **making the gate a required check** are honest, documented residuals with explicit deploy/follow-up steps. gate-manifest 70/70, prosecute 129/129, full suite 4739/4739, mutation-gate 20/20.

voodootikigod:
## Added: local `.env.local` → `ADLC_MANIFEST_KEY` loader (commit 9a1283b)

So `record-cross-model` / `tier-check` pick up the signing key from a gitignored `.env.local` without an explicit shell export. Because it touches a secret, the loader (`packages/prosecute/lib/load-env-local.mjs`) is deliberately narrow:

1. **Only `ADLC_MANIFEST_KEY`** is read from the file — no other var is ever injected.
2. **An already-set `process.env` value always wins** — never overrides/shadows a real key.
3. **Never loads under CI** (`GITHUB_ACTIONS` / `CI`), checked **first** — the trust-root gate runs with cwd = the PR-controlled tree, so reading a key out of a checked-out `.env.local` there would let a contributor supply their own key and forge an approve. Even an unconfigured (empty) secret can't fall through to the file.

Tests: parser + all three guard rules (unit) and CLI proofs that a local run signs/verifies via `.env.local` while a `GITHUB_ACTIONS` run ignores the file and fails closed, plus an exported key winning over the file. prosecute 146/146, full suite 4756/0, mutation-gate all-killed (incl. both `isCIEnv` comparisons). A focused codex review of the loader is running.

voodootikigod:
## Loader review round 3 — F2 (tree-supplied verification key) resolved

The reviewer confirms the fix: the loader now runs **only in `record-cross-model`** (signing), never `tier-check` (verifying), and explicitly endorses it — *"Retain the current `.env.local` process-boundary tests; they adequately pin signing-only loading and fail-closed verification."* Added the force-added-`.env.local` + forged-approve regression test (local `tier-check` fails closed). No new loader findings. Full suite 4758/0, mutation-gate 16/16.

The two HIGH findings this round are **not** loader issues — they're the tracked residuals:
- **Truncation** (`cross-model.mjs`) → issue #355 (drop a signed revocation; needs an external anti-rollback anchor).
- **"Gate is no longer a required merge check"** → the enforcement/deploy item.

### ⚠️ Enforcement transition (worth a conscious decision at merge)
Merging #354 **removes** the folded cross-model step from the required `test` job and replaces it with the separate `cross-model-gate.yml` (which is *not* a required check until branch protection says so). So:
- **Today on main:** the folded step is "required" but forgeable/keyless — weak enforcement.
- **After #354, before branch protection:** no required trust-root gate; the maintainer (P6) is the enforcer, same as everything else on this branch-protection-less repo.
- **After #354 + branch protection marking `cross-model-gate / gate` required + `ADLC_MANIFEST_KEY` set:** required **and** unforgeable (modulo the #355 truncation limit).

Net: not weaker than today's forgeable-required step in practice, but the *formal* required status is only restored once you configure branch protection. Deploy checklist stands: (1) set the secret, (2) mark the gate required.

PR #357, opened 2026-07-26
feat(herdr): Phase 3 — fleet observer bridge (t-herdr-8, t-herdr-9)

voodootikigod:
## Phase 3 — herdr fleet observer (t-herdr-8, t-herdr-9)

Makes the herdr plugin **observe a running `adlc fleet`** and reflect it into
herdr, read-only, regardless of coding harness. Delivers the two Phase-3 tickets:

- **t-herdr-8** — `fleet-status.json` becomes a **versioned, read-only observation
  surface**: `@adlc/fleet` stamps `FLEET_STATUS_SCHEMA_VERSION`, round-trips it
  through persistence, and the spec (`docs/specs/fleet-orchestration.md`) documents
  the field + the "unknown/absent version → degrade to polling `fleet status
  --json`" contract an external observer can depend on.
- **t-herdr-9** — a **pure planner + thin executor** in the plugin:
  `planFleetBridge(prev, curr, knownSchemaVersion, seenRunIds)` decides *what* to
  do (open a run tab once, tail in-flight ticket logs, notify on terminal
  transitions, summarize on the board); the daemon glue only executes. Untrusted
  input throughout: an unknown schema degrades to polling, and hostile ids never
  reach a path or argv.

### What the observer does
- Opens a `fleet: run-<id>` tab once per run; inside it, a `tail -F` pane per
  in-flight ticket (`-F`, not `-f`, so it waits for the log the orchestrator
  creates a beat later). Each pane gets a **per-ticket agent name** and a native
  herdr **token tag** with the ticket's current state, so concurrent panes are
  distinguishable.
- Notifies on terminal transitions (merged → done sound; failed/blocked → request
  sound, with the per-ticket worktree to inspect in the body).
- Summarizes terminal tickets on the board pane; a **failed/blocked ticket keeps
  its tail pane open** so its error output stays readable (merged releases its
  pane).
- Fails soft everywhere: every herdr effect is best-effort and logged (never
  swallowed); a bad/absent status degrades to polling; `planFleetBridge` is total
  on any untrusted input.

### Bounds & hygiene (all learned from the prosecution loop)
- Pane close, tail spawn, and token tag are each **bounded** so a
  manually-closed pane/tab can't drive a per-400ms `herdr` spawn loop; the tag
  refreshes on the 45s heartbeat exactly like the main token loop.
- Per-ticket spawn-failure counts are cleared on a new run (no cross-run
  poisoning); the transition baseline is preserved across a null/unreadable
  status read (no UI churn / lost scrollback).

### Known limitation (deferred to Phase 4)
The observer's pane-tracking state is **in-memory**. On a daemon restart mid-run
it opens a fresh tab and leaves the prior run's panes as harmless orphans. A
persisted-pane-id recovery was prototyped and **deliberately reverted**: herdr
reuses pane ids after a herdr-daemon restart, so a persisted id can match an
unrelated pane the user has since opened — adopting then closing it would destroy
their session. A sound fix needs herdr to expose a pane's agent/command in `api
snapshot`. Tracked as a Phase-4 follow-up (see `bin/watcher.mjs`).

### Prosecution
Built via the ADLC lifecycle and hardened through a cross-model
adversarial-review → fix → re-review loop (agy vs the Claude author): **~20 rounds**
that surfaced and fixed real defects — pane lifecycle leaks, notification storms,
IPC-failure isolation (herdr signals failure by `{ok:false}` return, not throw), a
`tail -f`-vs-`-F` race, a `typeof null === object` poison loop, missing pane
labels, and a CRITICAL data-loss risk in an early restart-recovery attempt (which
drove the revert above). Findings converged to minor UX/hygiene items.

### Test plan
- [x] `plugins/adlc-herdr/test/*.test.mjs` — 266 pass (fleet-bridge rail 42:
      planner, executor, all bounds, retention, argv builders; plus board,
      watcher, panemap, tokens, etc.)
- [x] `packages/fleet/test/status-schema.test.mjs` — schemaVersion + round-trip +
      spec contract
- [x] `npm run preflight` — all 5 PR-blocking gates green (run-tests, rails-guard,
      rail-freeze, mutation-gate, findings-ledger/append-only)
- [x] Cross-model adversarial-review (agy vs the Claude author) — **APPROVE, no
      material findings** on the full diff; P5 approve recorded to the manifest
      for t-herdr-8 + t-herdr-9

> Prosecution note: the review→fix loop ran ~24 rounds before converging to a
> clean APPROVE. The deferred daemon-restart recovery is tracked as a follow-up
> (see the "Known limitation" section) — it can't be made sound until herdr
> exposes a pane's agent/command in `api snapshot`.


voodootikigod:
Deferred daemon-restart recovery (the "Known limitation" section) is tracked as a Phase-4 follow-up: #358 — it can't be made sound until herdr exposes a pane's agent/command in `api snapshot`.

Issue #359, opened 2026-07-26
mutation-gate: off-by-one on tuning constants is an unkillable equivalent mutant, and the exemption the failure message advertises does not exist


voodootikigod:
## Symptom

`mutation-gate` fails on a mutant that **no test can ever kill**, and the remediation its own failure message prints does not exist.

Concrete instance, from #353 (`scripts/rails-guard-ci.mjs`, round 6):

```
SURVIVED  scripts/rails-guard-ci.mjs:53  [off-by-one]
  original: spawnSync('git', args, { encoding: 'utf8', timeout: 60000, maxBuffer: 512 * 1024 * 1024 });
  mutated:  spawnSync('git', args, { encoding: 'utf8', timeout: 60001, maxBuffer: 512 * 1024 * 1024 });
Total: 12  Killed: 11  Survived: 1
```

A 1 ms delta on a 60-second git timeout is not observable by any test. Writing one would require a test that actually spends 60 seconds and still could not distinguish 60000 from 60001. This is an equivalent mutant, not a coverage gap.

## Root cause — two separate defects

**1. `off-by-one` fires on tuning constants.** `packages/core/lib/mutate.mjs:146` takes the *first* number on the line and adds one. Combined with `mutate.mjs:237` (only diff-changed lines are targeted), **any** line that introduces or touches a timeout, buffer size, retry count, or limit becomes an unkillable target. This is structural: the constant is a tuning knob, so by construction no behavioral assertion pins its exact value.

Note the operator picks only the first number per line, so in the example above `maxBuffer: 512 * 1024 * 1024` was never mutated — the round-6 `>1 MiB` test (`scripts/test/rails-guard-ci.test.mjs:484`) does cover that behavior. Only the timeout survived.

**2. The advertised exemption is not implemented.** `scripts/mutation-gate.mjs:310` tells the operator:

> Either add a test that fails when that line is altered, or, if the line is genuinely redundant with a stronger check, say so in a comment AT ITS DEFINITION and re-run.

There is no comment-based exemption anywhere in `mutate.mjs`. `SKIP_LINE` (`mutate.mjs:223`) only skips lines that *begin* with `//`, `/*`, `*`, `#`, `import`, `export {`, or `console.`. A comment above or beside the line does nothing. Following the gate's own instructions verbatim cannot make it pass.

## Why the obvious workarounds fail

[...]

Issue #314, opened 2026-07-23
rails-guard-ci is unusable locally: a gitignored .adlc/manifest.jsonl triggers a false deny


voodootikigod:
Running the rail gate locally denies on an untracked, **gitignored** file, so the local check disagrees with CI on an unmodified working tree.

## Reproduction

```
$ git check-ignore -v .adlc/manifest.jsonl
.gitignore:25:.adlc/*	.adlc/manifest.jsonl

$ git log --oneline --all -- .adlc/manifest.jsonl | wc -l
0

$ node scripts/rails-guard-ci.mjs origin/main
rails-guard-ci: .adlc/manifest.jsonl cannot be created with evidence in a PR;
create it empty during bootstrap or use the protected-base runner ceremony
```

The file is ignored, is in zero commits, and is not in the branch diff — yet the gate denies. The same branch passes `rails-guard` in CI, because CI checks out clean and the file simply does not exist there.

## Why it matters

`ADLC_RAILS_BYPASS` exists so a deliberate override is *recorded*. A gate that cries wolf on an unmodified tree pushes people toward setting that variable habitually, which is exactly the erosion #162 was about — and the bypass is session-lifetime and unrevocable (#204), so the habit is expensive.

Concretely, this fired on every local run across this week's work. Each time I had to re-derive that it was spurious by checking `git check-ignore` and the commit list. Anyone less suspicious either bypasses or stops running the gate locally, and both defeat the point of having a local check.

The file is produced by ordinary `adlc` usage (`gate-manifest record`, and the ticket commands), so anyone who uses the toolkit in the repo will have one.

## Likely cause

The check appears to ask "does `.adlc/manifest.jsonl` exist in the working tree while being absent from the base ref?" and treat that as a PR creating it with evidence. On a clean CI checkout the answer is no; on any developer machine that has run the toolkit, the answer is yes. The distinction it actually wants is **"is this file being ADDED BY THE DIFF"**, which is a git question, not a filesystem one.

## Suggested fix

Decide the manifest's state from the **diff**, not from the filesystem: a file that is gitignored and untracked cannot be introduced by the PR, so it should never trigger this deny. The bootstrap-evidence check should look at what the commit adds, consistent with how the rest of the gate reads the base ref rather than the working tree.

## Acceptance sketch

- With an untracked, gitignored `.adlc/manifest.jsonl` present locally and no diff touching it, `node scripts/rails-guard-ci.mjs origin/main` exits 0.
[...]

PR #363, opened 2026-07-27
#140 — single-source the rails-guard gate as `adlc rails-guard-ci`

### LINKED ISSUE #140: rails-guard: single-source the gate, generate the inline CI template instead of hand-syncing + hashing
## Context

Surfaced dogfooding #104 (PR #135). The rails-guard gate logic lives in **two places**:

- `scripts/rails-guard-ci.mjs` — the real gate.
- `docs/ci/rails-guard.yml` — a self-contained **inline copy** of the same node script, embedded in the workflow YAML.

They're kept in sync by a sha256 fixture (`scripts/test/rails-guard-workflow-hashes.json`) that `rails-guard-bootstrap-ci.test.mjs` recomputes over the live template. Every gate change means **editing both files and regenerating the hash** — I did exactly that in #104 (the `isCompletionAnnotationOnly` relaxation had to be mirrored by hand, then the fixture regenerated).

The hash test *catches* drift, but doesn't *remove* the double-edit burden, and hand-mirroring a security-critical gate is precisely where a subtle divergence can slip in.

## Proposal

Make the `.mjs` the **single source of truth** and *generate* the inline YAML script from it:

- A build/CI step that inlines `rails-guard-ci.mjs` (or a bundled subset) into `docs/ci/rails-guard.yml`.
- The hash pin then asserts a much weaker, safer invariant — "the generated artifact is fresh" — instead of "two independently hand-edited copies happen to agree."
- Gate authors edit one file; the template is derived.

## Tradeoff / note

The inline template exists so the workflow is self-contained (no dependency on repo scripts at CI time). Generation preserves that (the YAML still ends up self-contained) while removing the manual sync. Needs care that the generator itself is trustworthy (it's touching a trust root) — a small, reviewed, well-tested inliner.

## Acceptance sketch

- Editing only `rails-guard-ci.mjs` and running the generator updates `docs/ci/rails-guard.yml` deterministically.
- The freshness check fails if the YAML is stale vs the `.mjs`.
- No hand-editing of the inline script is required for a gate change.

## Evidence

- #104 / PR #135 (relaxation mirrored by hand across script + template + hash fixture).

voodootikigod:
Closes #140.

## The issue's premise did not survive contact with the code

#140 proposed generating the inline `docs/ci/rails-guard.yml` script from `scripts/rails-guard-ci.mjs`. They were not two copies of one program — they were **different programs**:

| | `scripts/rails-guard-ci.mjs` | `docs/ci/rails-guard.yml` inline |
|---|---|---|
| deps | imports `@adlc/tickets`, `@adlc/rails-guard` | zero-dep `node -e` |
| #141 trust-root ceremony | yes | no |
| legacy→directory migration | verifies evidence, allows | denies (fails closed) |
| bootstrap ack / CODEOWNERS coverage / signed-runner probe | **no** | yes |
| manifest read | committed blob-by-hash, symlink-hardened (#314) | `existsSync`/`readFileSync` |

Neither was a subset of the other, so a literal inliner would have had to bundle two npm packages into a shell-quoted string *and* pick a winner on migration policy.

## What shipped instead

The gate ships as **`adlc rails-guard-ci`** in `@adlc/rails-guard`. This is not a new trust dependency: the template already installed the same pinned `@adlc/cli` and already delegated the verdict to `adlc rails-guard`. The pin now carries the gate too.

```
packages/rails-guard/lib/ci/          the gate, in focused modules
packages/rails-guard/bin/rails-guard-ci.mjs
docs/ci/rails-guard.yml               551 → 154 lines, no inline script
scripts/rails-guard-ci.mjs            686 → 76 lines, a wrapper declaring
                                      this repo's own extra trust roots
scripts/test/rails-guard-workflow-hashes.json   deleted
```

The sha256 fixture could prove the copies were UNCHANGED but never that they AGREED — and they had already drifted: **#314 fixed a manifest fail-open in the script that the template never received.**

## Behaviour changes (documented in the template header)

- A legacy→directory ticket-store **migration with valid hash-chained evidence is now allowed** for template users instead of denied. The old denial was not policy — the inline copy *could not verify* migration evidence, so it failed closed.
- The template inherits #314's committed-blob manifest reader (symlink, ancestor, TOCTOU hardening) and the #141 ceremony.
- Ticket stores are read via the canonical `@adlc/tickets` loader, so malformed/ambiguous/symlinked shards and filename↔id mismatches fail closed with that loader's diagnostics.
- Verdicts `throw` `GateDeny`/`GateFail` instead of calling `process.exit` from inside a helper, so a verdict cannot be silently fallen through.
- `rails-guard-workflow-hashes.json` is dropped from the gate's freeze list but **kept in prosecute's tier classifier**, where ticket T41 AC3 requires it.

## Repo-specific trust roots

The wrapper passes `--trust-root` for paths only this repo has (`.github/workflows/ci.yml`, `packages/rails-guard/lib/ci/**`, the verdict bin, …). That is safe on the command line **because the wrapper is itself a built-in default trust root** — a PR cannot shrink the list without editing a frozen file and tripping the ceremony. Verified: running the gate on this branch denies with exit 2 and names the new `lib/ci/**` files.

## Test plan

- [x] `scripts/test/rails-guard-ci.test.mjs` — **90/90 pass unchanged.** This was the parity harness for the rewrite.
- [x] `packages/rails-guard/test/*` — 216/216 (adds `ci-args`, `ci-trust-roots`, `ci-codeowners`, `ci-bin`).
- [x] `scripts/test/rails-guard-bootstrap-ci.test.mjs` — 80/80. Now runs the real bin instead of a script extracted from YAML; its five argv-capture tests became **verdict** assertions ("editing this path is denied"), which the PATH stub could no longer observe and which are the property that actually matters.
- [x] Full `npm test` — no regressions. Three unrelated failures: `board-e2e` fails identically on clean `main`; two `npm publish --dry-run` tests are network-dependent and pass in isolation.
- [x] `mutation-gate` — 10/11 killed.
- [x] `scripts/rails-guard-ci.mjs` — exit 2, correctly denying its own trust-root change.

## Expected red checks (per the solo-maintainer pattern)

1. **`rails-guard`** — ceremony needs a non-author CODEOWNER approval; structurally impossible solo.
2. **`gate` (cross-model)** — trust-root tier; needs a distinct-provider attestation.
3. **`mutation-gate`** — one survivor: `MAX_BUFFER_BYTES = 512 * 1024 * 1024` → `513`. An **equivalent mutant** — no test can observe 1 MiB of difference in a spawn buffer bound. Killing it would require a source-text pin, i.e. the hollow test #359/#362 exists to eliminate. PR #362 is the systemic fix for this operator.


voodootikigod:
## Cross-model review (P5) — round 1: NO-SHIP → fixed in `f0e5fbc`

A distinct-provider adversarial review (codex, read-only) returned **NO-SHIP** with two blocking findings. Both were verified against the code, not taken on trust, and both are fixed with a regression test.

### 1. The trust-root list could mutually unprotect itself

The gate's own sources were passed in from `scripts/rails-guard-ci.mjs` via `--trust-root`, on my argument that *"the wrapper is itself a default trust root, so the list cannot shrink."* **That argument was wrong.** One PR could delete the wrapper's arguments **and** remove the wrapper from `DEFAULT_IMMUTABLE_TRUST_ROOTS` — lifting both enforcement sources in a single change, with no ticket rail covering either.

Fix: the gate sources live in the **defaults** now, so disarming the gate means editing a file the list already covers. The wrapper passes only `.github/workflows/ci.yml`, which is genuinely repo-local.

The residual is irreducible and pre-existing: a same-repo `pull_request` runs the PR branch's gate, so a PR editing `trust-roots.mjs` runs its own edited list. Nothing in-repo closes that — branch protection is the backstop. It is now documented *at the list* instead of implied by a two-file indirection.

### 2. A trailing-slash CODEOWNERS pattern falsely reported coverage

`codeownersMatch` built `^rx(?:/.*)?$` for a directory pattern. The **optional** suffix let `docs/` match `docs`, and let `.github/workflows/adlc-rails-guard.yml/ @owner` "own" the like-named **file**. GitHub treats a trailing slash as a directory and leaves that file ownerless — and the old inline template agreed, because it appended `/**`.

Verified directly:

| pattern vs `.github/workflows/adlc-rails-guard.yml` | new (before fix) | old template |
|---|---|---|
| `.github/workflows/adlc-rails-guard.yml/` | `true` | `false` |

This is a **latent bug on `main`**, but #140 made it *reachable* from the bootstrap self-protection check — turning it into a fail-open where a repo could self-attest that its deployed workflow was protected when it was not. The directory form is now exact (`^rx/.*$`).

The pre-existing comment argued over-matching is fail-safe because "a wider owner set only makes approval harder to forge." True for owner **resolution**; false for **coverage**, which is a yes/no question. That distinction is now written at the function.

### Confirmed NOT problems (review checked explicitly)

- **No verdict is swallowed** by the `process.exit` → `throw GateDeny/GateFail` restructure. Ticket-store failures propagate, the signed-runner `finally` suppresses only cleanup errors, and every `fail()` call site remains non-returning.
- **No old trust root was lost** except the intentionally deleted hash fixture.
- `packages/rails-guard/lib/ci/**` works as a git pathspec for additions, nested paths, and renames (independently verified).
- CODEOWNERS file precedence (`.github` → root → `docs`, first-existing wins) matches both the old template and GitHub.

### After the fix

217 rails-guard package tests, 178 gate/bootstrap tests, mutation-gate 10/11 (same equivalent-mutant survivor). The gate still self-denies its own trust-root change — exit 2, naming all 9 `lib/ci` files.

Issue #370, opened 2026-07-27
record-cross-model silently writes an UNSIGNED attestation and reports success — the gate rejects it, after a full review has been spent


voodootikigod:
## `record-cross-model` reports success for an attestation the gate will reject

With no `ADLC_MANIFEST_KEY` in the environment, `record-cross-model` writes an **unsigned** entry, exits **0**, prints a success line, and emits **nothing** on stderr.

Verified directly (scratch repo, `ADLC_MANIFEST_KEY=''`):

```
exit   : 0
stdout : "recorded cross-model approve for T1 @ git-worktree:deadbeef (agy vs author anthropic)\n"
stderr : ""
entry signed: false
```

An unsigned approve does not satisfy the trust-root gate — `#326`'s forge-resistance test pins exactly that. So the operator receives explicit confirmation of success for evidence that is inert.

## Why this is expensive, not just untidy

The attestation is the **last** step of the ceremony. The sequence is: run a distinct-provider adversarial review → record the result → push → gate. So the operator:

1. Spends a full cross-model review (real time, real spend).
2. Records it, and is told it worked.
3. Pushes.
4. The gate stays red with the *missing attestation* message — pointing back at the step they just successfully completed.

**The wasted review is the cost**, and it is the same failure class as #364: a gate surface that does not say what actually happened. #364 is arguably milder — there, `record-cross-model` at least *errored* and named the true cause. Here nothing anywhere reports a problem until CI.

There is a second, permanent cost: `.adlc/manifest.jsonl` is append-only and hash-chained. An unsigned entry committed by mistake **cannot be removed** without breaking the chain, so it is noise in the ledger forever.

## How this was hit for real

While recording the attestation for #367. A key-extraction command silently produced an empty string (`\+` / `\?` are GNU `sed` extensions that BSD `sed` on macOS does not support), so the key never reached the process. `record-cross-model` reported success. It was caught only by inspecting the written entry rather than trusting the success message, and discarded while still uncommitted — had it been committed it would be permanent.

A maintainer who trusts the success line — the reasonable thing to do — gets no signal at all.

## Note: the unsigned path is deliberate, and must stay reachable

`#326`'s forge-resistance test records an unsigned approve on purpose, to prove the gate rejects it. So this cannot simply be made a hard failure without updating that test. The fix needs to distinguish an **operator who lost their key** from a **test (or attacker) deliberately writing unsigned**.

## Proposed work

[...]

PR #379, opened 2026-07-28
ADR-0015 + spec: segmented gate-manifest (chain forest), with P0 ticket DAG

voodootikigod:
## Problem

`.adlc/manifest.jsonl` is a single hash-chained ledger — every entry carries
`seq`, `prev` (sha256 of the previous raw line), and an HMAC `sig` (v2 signs
all fields including `seq`/`prev`). Every gate appends through the same file,
so two concurrent branches always collide at the tail: a guaranteed merge
conflict whose resolution requires rewriting `seq`/`prev`, which invalidates
signatures and needs the signing key to fix. Evidence: commits 56de76a,
04bfb7b, 8c0fcdc, and two `manifest.jsonl.pre-repair-*.bak` files in this
repo's history.

## Proposal

ADR-0015 + normative spec: freeze the current file as an immutable root
segment; new evidence appends to per-lineage segment files under
`.adlc/manifest.d/`, each bound into committed history by a fork anchor.
Forest-aware readers ship a release before writers do. Greenfield repos
scaffold directly on the new format; this repo migrates separately under
`T-MANIFEST-FOREST-MIGRATE`. Precedent: ADR-0010 / `T-SHARDED-STORE` /
migration PR #321.

## Ticket DAG (P0 output)

`T-MANIFEST-FOREST-SPEC` (this PR, contract: the spec) →
`T-MANIFEST-FOREST` (implementation, trust-root tier) →
`T-MANIFEST-FOREST-MIGRATE` (this repo's cutover, separate PR).
Read with `adlc ticket show <id> --json`.

## Gates run

- `adlc spec-lint docs/specs/segmented-gate-manifest.md` — 11/11 exit 0.
- Premortem (self-answered in-session, prior to this PR) found and fixed 4
  defects in the original draft: lineage-token branch binding, anchor
  resolution at HEAD not base, activation marker location, empty-slug
  fallback.
- Adversarial design review (self-answered, prior to this PR) found and fixed
  2 more: unsigned revocations counting fail-closed (§6), cutover-entry
  spoofing named and shown self-denying (§9.1).
- **`npx adversarial-review --base main`, run fresh at PR time (five passes,
  local `agy` CLI, cross-model)** — this is the "fresh cross-model scrutiny"
  the premortem/design-review caveat said was still owed. It found real gaps
  the self-answered rounds missed and each was fixed before the next pass:
  1. §7.1's branch-mismatch anchor fallback could anchor a new segment to a
     segment that existed on disk (segments aren't gitignored) but was never
     committed on the current branch — dangling anchor, denied at CI §9.3.
  2. The narrower fix for #1 still fell through to `anchor: null` in a
     root-less repo where *other* lineages already had committed segments —
     which §9.3 denied once any segment existed.
  3. §6's terminal-revocation rule has no exception for cutover seal entries,
     so a migration seals every standing approval as revoked for *all*
     readers, forever — true, and deliberate (D1: migration forces fresh
     re-approval under the forest's trust semantics), but nothing in the spec
     said so out loud, so it kept getting flagged as a defect. Added the
     rationale inline at §4.6.
  4. Two tickets (`T-MANIFEST-FOREST`, `T-MANIFEST-FOREST-MIGRATE`) still
     described the migration ceremony as *refusing* to migrate while
     approvals stand, from before the seal design was finalized. Corrected
     via `adlc ticket update --write` to match §4.6/§8.
  5. The real bug under #1/#2: two branches forked from the same root-less
     state each legitimately mint an independent `anchor: null` segment
     (§5's forest is explicitly rooted in root *or* `null` segments, plural)
     without coordinating — whichever merged second was denied by §9.3's
     stricter "no root **and** no segments" null-anchor rule, and its already
     -signed anchor can't be repointed after the fact. Relaxed §9.3 to key
     off "no root" only; simplified §7.1's fallback back down to
     root-or-null (the fallback-chasing logic from fix #2 turned out to be
     unnecessary complexity once #5 was fixed correctly). Updated AC8 to
     match.
  - The two findings still open in the final pass (unsigned revocations as a
    DoS vector on approvals; "anchors don't cryptographically bind sibling
    segments") are the same tradeoff surfacing twice more — both are argued
    explicitly inline in the spec (§6's fail-closed rationale, §9.3's
    "history pinning comes from append-only, not from anchors"). Not new;
    included here for the human reviewer to weigh, not silently dropped.

## What a human reviewer should actually decide

ADR status is **Proposed**. The two points above (unsigned terminal
revocation as an accepted DoS surface; anchors as bookkeeping rather than a
security boundary) are the load-bearing judgment calls in this design — flip
the ADR to Accepted only after agreeing with those, not just the mechanics.

## Test plan

- [ ] Human reads ADR-0015 and the spec, and either accepts the two tradeoffs
      above or asks for a redesign before flipping Status to Accepted.
- [ ] `adlc spec-lint docs/specs/segmented-gate-manifest.md` exits 0 (already
      verified, re-run if the spec changes further in review).
- [ ] After merge: `T-MANIFEST-FOREST` becomes coldstart-executable; next
      session starts implementation, readers before writers per spec §11.


voodootikigod:
Rebased onto current `main` (picked up #370/#236/#376/#377, no conflicts) and resolved the `gate` (cross-model) check: this branch is trust-root tier (touches the ADR/spec docs, which `T-MANIFEST-FOREST`'s own rails freeze once implementation starts), so it needs a distinct-provider cross-model approve bound to the PR revision.

Ran two more rounds of `npx adversarial-review` against the rebased base (agy CLI, distinct from this PR's anthropic author) — found and fixed 5 more real spec gaps beyond the original 4 documented in the PR description:

- §4.4 still said `anchor: null` requires *no other segment to exist*, contradicting the §7.1/§9.3 fix from the prior round that explicitly allows it in root-less multi-lineage repos — reconciled.
- §8 step 4's migration-ceremony refusal check only looked at the `.store.json` marker, not §4.7's full marker-OR-cutover-entry test — the whole point of the double marker is redundancy against exactly that.
- §6's terminal-revocation match was described as a flat `(provider, revision)` tuple, undocumenting the optional `ticket` scoping that `candidateReview` (and §4.6's seals) actually rely on — spelled out both gate modes.
- §4.2 left ULID casing unspecified in the segment filename grammar; pinned to uppercase, matching this repo's existing ticket-id convention.
- ADR §6 undersold segment volume from automated CI recorders (ephemeral checkout, gitignored `.lineage`) vs. interactive branch lineages — noted, not a defect, but relevant to the compaction trigger.

Recorded the signed distinct-provider attestation (`agy` vs. `anthropic`) against the final rebased revision. `tier-check` PASSes locally and in CI now.

Two categories of finding surfaced in the last two passes were **not** changed, deliberately:
- Unsigned terminal revocations as an accepted DoS surface, and "anchors are bookkeeping, not a trust boundary" — both argued explicitly inline in the spec already (§6, §9.3); the reviewer re-flags them each pass but they're intentional trade-offs, not gaps.
- A per-ticket cross-model query excluding an un-ticketed (revision-wide) `needs-attention` entry — real, but it's **existing** behavior in `packages/prosecute/lib/cross-model.mjs` today, unrelated to this spec's forest change (which explicitly preserves matching "exactly as today"). Worth its own ticket if it's actually a live gap, but out of scope here.

Issue #378, opened 2026-07-28
Stop hook always reports evidence-chain INVALID: verify() doesn't tolerate legacy unsigned manifest history


voodootikigod:
## Problem

The Claude Code Stop hook (`plugins/adlc-claude-code/hooks/adlc-hook.mjs`, `manifest()`) runs `gate-manifest verify --json` after every session and surfaces a system message whenever the result is invalid:

> ADLC gate-manifest: evidence chain INVALID — chain broken at seq 1 (line 1): unsigned entry. The gate ledger may have been tampered with or truncated.

This fires in **every session**, not just when something is actually wrong.

## Root cause

`.adlc/manifest.jsonl` predates HMAC signing. Signing was introduced partway through the ledger's life — the first signed entry is `seq 96`; `seq 1`-`95` were written before `ADLC_MANIFEST_KEY` existed and are legitimately unsigned.

`packages/gate-manifest/bin/gate-manifest.mjs` calls `verify(flags.dir)` with no options, so it always runs with the default `requireSignatures: true`. Whenever `ADLC_MANIFEST_KEY` is present in the environment, that mode requires *every* entry — including the honest pre-signing prefix — to carry a valid signature, and fails closed at `seq 1` every time.

`packages/gate-manifest/lib/verify.mjs` already implements and documents the fix for exactly this case:

> `requireSignatures:false` → tolerate entries with NO sig (legacy/unsigned history) ... A PRESENT-but-invalid sig is STILL rejected (tampering), so this is "unsigned history is OK", not "tampered signatures are OK".

but the CLI (`gate-manifest verify`) never exposes or passes that flag, so the lenient path this repo actually needs is unreachable from the CLI or the Stop hook.

## Impact

- The Stop hook cries wolf on every single session, which trains users to ignore it — defeating its purpose as a tamper-detection signal.
- It gives no actionable signal today: a real forged/truncated chain would produce the *same* message as this permanent false positive.

## Fix

Plumb a way for `gate-manifest verify` (and the Stop hook that calls it) to accept the repo's known-legitimate unsigned prefix, without weakening real tamper detection:

- Add a flag (e.g. `--allow-legacy-unsigned`) to `gate-manifest verify` that passes `requireSignatures: false` to `verify()`.
- Any entry with a **present but invalid** signature must still fail verification in this mode — only *missing* signatures on legacy entries are tolerated. This is already `verify.mjs`'s documented and implemented behavior; it just isn't reachable from the CLI.
[...]

PR #384, opened 2026-07-28
feat(gate-manifest): forest read layer — T-MANIFEST-FOREST slice 1

voodootikigod:
## Summary

T-MANIFEST-FOREST slice 1 (readers before writers, per spec §11 / ADR-0015 decision 5): the chain-forest **read** layer. No writer changes — a segment-free repo (every repo today, including this one) verifies/reads byte-identical to before, plus a new `segments: n` field.

## What's here

- **`packages/gate-manifest/lib/forest.mjs`** (new) — segment discovery (§4.2 grammar, symlink/nested-dir/case-collision rejection), fork-anchor resolution (§5.4), anchor-cycle detection (§5.5), and `readManifestForest` — the `readEntries('manifest')`-equivalent forest read (§6), root then segments in a **topological** order (parent segments before anything anchored to them, spec's 3-key tuple breaking ties among siblings).
- **`packages/gate-manifest/lib/verify.mjs`** — refactored so the exact per-chain algorithm (byte-for-byte unchanged) runs once over root and once per segment. A segment-free repo takes an explicit fast path (AC2).
- Wired into gate-manifest's own `show`/`attest`, plus three `readEntries('manifest')` consumers whose "latest wins" logic is array-position-based (runner's P5-completion lookups, prosecute's P5-finding lifecycle tracker) — safe once given a deterministically ordered read.

## Deliberately NOT touched (documented inline, not silently dropped)

- **`prosecute/lib/cross-model.mjs`** (2 call sites) — spec §6's terminal-revocation rule ("no entry anywhere in the forest", unconditional) is a genuine strengthening over today's "latest per provider wins" reduction, not a drop-in read swap. That redesign is this ticket's own **slice 2**.
- **`ticket-sync/lib/outcomes.mjs`**'s `reduceTicketOutcomes` — picks "latest per gate" by comparing raw `seq` **values**, which are only meaningful within one chain; across segments (each restarting seq at 1) that's actively wrong, not just stale. Needs its own fix (compare by array position instead) that also has to preserve an existing, deliberately order-independent `ts`-fallback test. Not named in the ticket body — flagging rather than scope-creeping a fix in here.

## Process notes worth knowing about

- This worktree never had `npm install` run, so `@adlc/gate-manifest` was resolving through Node's upward module search into the **main checkout's unmodified copy** — meaning my first several test runs against `packages/prosecute`/`packages/runner` weren't exercising my changes at all. Caught via a hard `ERR_MODULE_NOT_FOUND` on the new `forest.mjs`; running `npm install` in the worktree fixed the symlinks. Worth remembering for future worktree sessions.
- Found and fixed an editing slip mid-session: a comment-only edit to `cross-model.mjs` accidentally deleted the `manifestChainTrustworthy` guard from `hasCrossModelApprove`. Caught by the full test suite (`cross-model.test.mjs`'s tamper-fail-closed test), not by review — a reminder that "just adding a comment" edits still need a full test run.
- Six rounds of `npx adversarial-review --base main` (agy, distinct from this session's anthropic authorship) found and fixed five real bugs beyond design: an unhandled `ENOTDIR` crash when `manifest.d/` is a regular file, a `signed:false` false-negative when a root-less repo's empty root chain forced `allSigned` permanently false, a non-topological sort that could order a child segment before its parent, an empty segment file vacuously passing the required-anchor check, and an unhandled `EACCES` in `readdirSync`. The final round's single finding was a hallucination (claimed a `+ 1` was missing from code that already has it) — verified directly against the file and left unchanged.

## Test plan

- [x] `packages/gate-manifest/test/forest-format.test.mjs` (new) — segment-free fast path byte-identical, valid multi-segment/root-less-concurrent forests, topological ordering, full AC1 adversarial rejection list.
- [x] Full repo suite: `npm test` — 52/52 segments green.
- [x] `npx adversarial-review --base main` — approve, no findings (6th pass).
- [x] `adlc-prosecute tier-check` — PASS (signed distinct-provider attestation, agy vs anthropic).


voodootikigod:
All checks green now. What happened between the first push and this:

**Main moved twice while this PR was in flight** — first #380 (unrelated docs), then #382 (`git-worktree:` → `git-change:` revision scheme, landing literally while CI was running). Rebased twice; the second rebase hit a real content conflict in `prosecute/lib/run.mjs` (main renamed `resolveRevision`→`resolveChangeSetRevision`) resolved by keeping the rename alongside this branch's `readManifestForest` swap. Both rebases also hit the manifest's own append-only structural-merge-conflict — ironic, since fixing that conflict class for real is this exact ticket's job on a repo (this one) that hasn't migrated yet. Resolved each by taking `main`'s manifest and re-recording fresh attestations, matching this repo's established pattern.

**`mutation-gate` caught 5 real gaps** the test suite's line coverage missed:
- `anchorOnFirst`'s `= false` default was dead code (every call site passes it explicitly) — removed rather than tested.
- `SEGMENT_NAME_RE`'s digit range was only ever exercised by digit-free slugs.
- `.lineage` being skipped (not merely "not invalid," but genuinely absent from both lists) had no assertion.
- `show.mjs`'s §6 segment-label line had no test at all.

All fixed with real tests (verified locally against `scripts/mutation-gate.mjs` before pushing — 6/6 killed).

voodootikigod:
Resolved the merge conflict — main moved again while this was open, this time landing a real content conflict (not just the manifest's own structural one):

**#378** (merged as `01614fb`, before that as `#383`) reworked `verify.mjs`'s signature checking: `requireSignatures:false` now tolerates unsigned entries only on a **contiguous legacy prefix** (before the first signed entry), not anywhere — closing a real regression where an attacker could append plain entries after a signed tail. Merged that logic into `verifyChain`, scoped **per-chain**: each call (root, or each segment) tracks its own local `seenSignedEntry` flag. In practice only root has a genuine legacy-unsigned prefix (this repo's own history predates signing; segments are new), but per-chain is the correct rule regardless. Added two tests neither side's original PR covered: a root with a real legacy prefix alongside a fully-signed segment, and an unsigned entry appended after root's own signed era (proves the anti-regression guard survived the merge).

Also resolved the manifest's own structural conflict again (3rd time this branch has hit it) the same way as before.

**Two more `adversarial-review` rounds on the rebased branch found 4 more real bugs:**
- `discoverSegments` checked `existsSync(manifest.d/)` before `lstatSync` — but `existsSync` follows symlinks, so a *dangling* symlink (target doesn't currently exist) returns false and skips the symlink check entirely, the exact case it exists for.
- `verify()`'s segment-free fast path returned `break` without the `segment: 'root'` label every other failure path includes — an API contract gap for callers inspecting `break.segment`.
- The segment sort's ULID tiebreaker wasn't total: two different segments (different slugs) sharing a ULID both compare equal. Not exploitable today (stable sort preserves `discoverSegments`' own deterministic order), but fragile to rely on — now falls back to the full filename.
- `parseLenient` didn't validate JSON.parse's result was an object; a line containing valid-but-non-object JSON (`null`, a number, a bare array) silently became a phantom `{segment: ...}` entry instead of being routed to `skipped`.

All fixed with regression tests, verified against `scripts/mutation-gate.mjs` (6/6 killed) before each push. All checks green now except one long-running `ticket-store-platform` matrix job still finishing.

PR #394, opened 2026-07-29
fix(install): Antigravity install fails — agy parses the plugin path as plugin@marketplace

voodootikigod:
## The bug

`curl -fsSL https://www.agenticlifecycle.ai/install.sh | sh` reported:

```
  > Google Antigravity detected
added 1 package in 465ms
Error: unknown marketplace: adlc/antigravity
  ! Google Antigravity: agy plugin install failed
```

`agy plugin install <target>` resolves its argument as `plugin@marketplace`
**before** deciding whether it is a filesystem path, so an `@` anywhere in the
argument is read as that separator. Every location npm gives a scoped package
contains one, so handing agy the real directory splits
`.../node_modules/@adlc/antigravity` into plugin `.../node_modules/` and
marketplace `adlc/antigravity`, and agy never touches the disk.

Reproduced directly against agy 1.1.8:

```
$ agy plugin install /tmp/nope@somemarket/someplugin
Error: unknown marketplace: somemarket/someplugin
$ agy plugin install /tmp/nope-plain
Error: install target must be a directory: /tmp/nope-plain
```

## The fix

agy **copies** the plugin into `~/.gemini/config/plugins/<name>/`, so a throwaway
directory is a complete install source. `install.sh` and the plugin's own
`adlc-agy` helper now copy the package to an `@`-free staging path, install from
there, and discard it.

Two further defects surfaced by adversarial review of the first commit:

- **`npx adlc-agy install` names a package that does not exist.** `adlc-agy` is a
  *bin* inside `@adlc/antigravity`, so npx only resolves it on a machine that
  already has the global install; anywhere else it hits the registry and 404s
  (verified against a clean npm prefix). The name is also unclaimed on npm, so
  every doc pointing at it aimed users at a name someone else could publish
  under. All surfaces now say `npx @adlc/antigravity install`.
- **Staging was only `@`-free by luck.** GNU `mktemp` and `os.tmpdir()` both
  honour `TMPDIR`, so `TMPDIR=/var/tmp/user@example.com` put the `@` straight
  back into agy's argument. Both installers now detect that and retry under
  `/tmp`, failing loudly if no `@`-free root exists.

The same broken command was published as the recommended manual step in
install.sh's summary, `docs/integrations/antigravity.md`, the plugin README, the
docs-site MDX page, and the marketing integration facts — all corrected.

## Verification

- Every new test was confirmed to **fail** against the unfixed code before being
  kept (4 tests: `@`-free target and hostile-`TMPDIR` staging, for both
  `install.sh` and `bin/cli.mjs`).
- The install.sh hostile-TMPDIR test **shims mktemp** rather than exporting a
  hostile `TMPDIR`: BSD mktemp (macOS) ignores `TMPDIR` for a template-less
  `-d`, which would have made the test vacuous locally while the bug stayed live
  on Linux CI.
- End-to-end against a real agy 1.1.8 from the real npm-scoped global location:
  installs cleanly, staging directory removed.
- Full suite: **52/52 segments green**.
- `npx adversarial-review --base origin/main --provider codex --verify`: APPROVE,
  no material findings.

## ⚠️ Merge blocker: `install.sh` is a frozen rail

`npm run preflight` fails the rail-freeze gate:

```
rails-guard: 1 violation(s) found
  [rail-edit]   apps/docs/public/install.sh
```

Ticket **I2** ("First-party one-command installer") freezes
`apps/docs/public/install.sh` and is still **open**, even though the installer
shipped in #351. The gate is designed so a PR can *never* unfreeze a railed
ticket (`isCompletionAnnotationOnly` only permits `completed: true` on tickets
that froze **no** rails), so this cannot be resolved from inside this PR.

**I2 needs to be marked `completed: true` on `main` before this can merge.** No
other file in this diff is railed — install.sh is the only violation.

## Test plan

- [ ] I2 marked completed on main, then `npm run preflight` passes
- [ ] CI green
- [ ] After release: `curl -fsSL https://www.agenticlifecycle.ai/install.sh | sh`
      on a machine with `agy` reports `installed for: ... Google Antigravity`


voodootikigod:
Rail-freeze blocker prepared: `chore/complete-i2` (commit `b34c956`) marks ticket I2 completed via the canonical ceremony (`adlc ticket complete I2 --write --authorize`), recording signed append-only manifest evidence. The annotation is add-only — rails, scope, edges and every other field are byte-identical to base.

Verified locally that it releases the rail: with this PR rebased onto that commit, `node scripts/rails-guard-ci.mjs` reports `rails-guard: all checks passed` (against current `origin/main` it reports the `[rail-edit] apps/docs/public/install.sh` violation).

Because completing a **railed** ticket is deliberately impossible from a PR (`isCompletionAnnotationOnly` accepts `completed: true` only on tickets that froze no rails — see `packages/rails-guard/lib/ci/rail-freeze.mjs:96`), `chore/complete-i2` needs to land on `main` as a protected-base admin action, not as an ordinary merge. Once it is on main, this PR needs a rebase and it should go green.

voodootikigod:
### CI triage — 4 failures, 1 fixed, 1 pre-existing, 2 blocked on I2

**`mutation-gate` — FIXED (`6d429c8`), now green in CI.**
Two survivors on the cleanup line inside the `catch`: `recursive: true→false` and `if (stage)→if (!stage)`. The cause was shape, not a missing assertion — cleanup was duplicated across a `catch` (staging failed) and a `finally` (install finished), and the catch copy is unreachable from any realistic test (`cpSync` copies FIFOs happily; a permissions trigger is root-dependent). Collapsed both into one `finally`, so every success-path test exercises the cleanup. Both mutants now die against the tests that **already existed** — no test was bolted on. Verified by simulating each mutation: baseline 2 pass/0 fail, each mutant 0 pass/2 fail. Local gate: `Survived: 0 — All mutants killed`.

**`test (18)` — pre-existing flake, not from this diff.**
`packages/prosecute/test/prosecute-tier-check-cli.test.mjs:594` fails its own **precondition** (`assert.equal(revA, revB)`), so it is a fixture bug, not a product failure. Revisions are `git-change:<commitSha>:<contentHash>`; in every failure the content hash is identical (`caf5359f…`) and only the commit SHA differs. `scratchRepo` (line 42) pins `user.email`/`user.name`/`gpgsign` but not `GIT_AUTHOR_DATE`/`GIT_COMMITTER_DATE`, and commit SHAs embed both at 1-second resolution — so the two repos collide only when both commits land in the same wall-clock second. Reproduced locally at ~50% (2/3, then 1/4) on Node 22; 18 vs 20/22 in CI is timing coincidence. Being run down separately.

**`rails-guard` — blocked on I2, resolves as soon as `chore/complete-i2` lands on main.**
Verified: rebased onto that commit, `rails-guard: all checks passed`.

**`gate` (cross-model) — blocked on I2, but will NOT clear by itself. Correction to my earlier note.**
I predicted completing I2 would clear this too. It does not. `rail-freeze.mjs:102` skips completed tickets (`if (ticket.completed === true) continue`), but `classifyTrustRootTier` (`packages/prosecute/lib/tier.mjs:122-129`) iterates every ticket's rails with **no `completed` filter**. Tested directly: rebased onto `chore/complete-i2`, tier-check still reports `touches rails deny-path of ticket I2` and exits 2.

Net effect worth a separate look: completing a ticket frees its rails for editing but leaves those paths permanently in trust-root tier — a ratchet that never releases. (Arguably `install.sh` *should* be trust-root tier on its own merits as a `curl | sh` target, but if so it belongs in `TRUST_ROOT_FILES`, not as a side effect of a closed ticket's rails.)

So `gate` needs a signed distinct-provider cross-model attestation regardless. The revision is base-dependent (`git-change:<baseSha>:<contentHash>`), so it changes when I2 lands and this PR rebases — recording one now would attest the wrong revision. Correct order: land `chore/complete-i2` → rebase this PR → re-run the distinct-provider review → record the attestation for the final revision.

voodootikigod:
Ceremony branch now has a PR for the review record: #397. Note it is **expected red** on `rails-guard` (exit 2, `base ticket I2 contract cannot change in the .adlc/tickets/ store in a PR`) — that is the intended forge-resistance, so #397 needs a protected-base push / admin merge rather than an ordinary merge.

voodootikigod:
### Round-16 review: 1 fixed, 2 declined with reasons

**Fixed — cancellation ordering (MEDIUM, conf 0.98).** The exit handler deferred settlement only on the TIMEOUT flag. An interrupt doesn't set it, so a leader exiting on forwarded SIGTERM settled the promise, the caller ran its `finally`, and the staged source was deleted while a SIGTERM-ignoring worker was still copying out of it — two seconds before escalation reached it. Timeout and cancellation now share one rule: neither is settled by the leader's exit. Test asserts *ordering* (the worker reports if its source vanishes underneath it), verified load-bearing.

**Declined — `/adlc-init` executes a PATH-resolved `adlc` (HIGH, conf 0.91).** Real class, wrong scope for this PR. `adlc init` appears in **18 files across every harness's docs** (codex, cursor, copilot, pi, antigravity, plus the integration guides). Hardening only the Antigravity copy would leave the other five inconsistent while implying the problem is solved. This is a repo-wide docs/policy decision about how ADLC tells users to invoke their globally-installed CLI — worth a ticket, not a unilateral edit inside an Antigravity bugfix.

**Declined — direct-copy fallback is uninterruptible during blocking fs calls (MEDIUM, conf 0.94).** True (`cpSync` is synchronous, so no handler runs during it) but the blast radius is bounded, and I verified this rather than assuming: the copy targets a **sibling** `incoming` directory and the live plugin is untouched until the two renames. An interrupted copy therefore leaves the live install fully intact plus a stale `incoming` sibling, which the next run removes unconditionally. The proposed fix is an async-fs rewrite of the fallback path; that is disproportionate to a leftover temp directory, and this branch has already shown that speculative rewrites of this subsystem generate more defects than they remove.

Both declines are recorded here rather than silently absorbed. Happy to file them as issues if you'd like them tracked.

PR #443, opened 2026-08-02
chore(adlc): complete T151 — quartermaster registry + routeJob (shipped in #402)

voodootikigod:
Marks **T151** completed now that #402 has merged.

## Why this is its own PR

This is the step that **expires T151's build-time rails** (T36 rails completion lifecycle). `rail-freeze.mjs` unions rails across *incomplete* tickets only, so while T151 sat open, **every PR in the repo** was frozen out of:

- `docs/specs/operating-stack.md`
- `packages/model-router/lib/assign.mjs`

Verified against the store: with T151 completed, no incomplete ticket declares either path.

A ticket-declared rail can never be *amended*, only **outlived**. This unblocks **#401** (ladder escalation), whose acceptance criteria include updating §5's escalation row and resolving §11 open question 4 — both inside the frozen spec. **#396** needs the same file next.

## T151's acceptance criteria are met by #402

| AC | Evidence |
|---|---|
| AC1 isolation | `packages/quartermaster/test/registry-isolation.test.mjs` |
| AC2 validation | `packages/quartermaster/test/registry-validation.test.mjs` |
| AC3 derivation | `packages/quartermaster/test/route-job.test.mjs` |
| Rail held: `assign.mjs` UNCHANGED | byte-identical to its state at #390, where the ticket was authored |

108 tests green across the quartermaster segment.

**Escalation and load-bearing transport were never in T151's scope.** They were carved out as #401 and #396 during its review, and it shipped with those limits documented rather than fixed. Nothing is being quietly waved through here.

## ⚠️ Expected red: `rails-guard`

`isCompletionAnnotationOnly()` deliberately **refuses** to credit a completion annotation on a ticket that declared rails:

> "the only effect of `completed` is to expire that ticket's rails, so annotating a RAILS-LESS ticket grants ZERO privilege — nothing to unfreeze... a railed ticket stays denied, so T36's forge resistance is fully intact for anything that could unfreeze a path."

That is the control working as designed: a PR that could complete a railed ticket could authorize its own rail edit. No ceremony lifts it — #424 (complete T152) took exactly this path.

## Verification

- [x] `adlc ticket complete T151 --write --authorize` — `sensitive: ["lifecycle-change"]`, `evidenceRequired: true`, `validation: valid`
- [x] Completion evidence **signed** at seq 158 (checked for `sig`, not assumed)
- [x] `gate-manifest verify --allow-legacy-unsigned` → `manifest ok (158 entries)`
[...]

Issue #396, opened 2026-07-29
Registry transport is validated and recorded but not load-bearing at dispatch — frontier and frontier-metered execute identically


voodootikigod:
## Summary

The quartermaster registry (T151) makes each seat declare a `transport` — `subscription:*`, `api:*`, or `gateway:*` — and validation treats that prefix as load-bearing. **Dispatch never uses it.** `packages/fleet/lib/live-deps.mjs` passes the seat's `adapter` and `model` to the adapter and drops `transport` entirely, so two seats differing only by transport execute identically and the harness picks whatever ambient auth it finds.

Surfaced by an adversarial-review round on the T151 branch. T151 ships with the limit **documented** rather than fixed; this issue tracks making transport load-bearing.

## Why it matters

The reference registry in `docs/integrations/quartermaster-registry.md` is exactly the shape that breaks:

```jsonc
"frontier":         { "adapter": "claude-code", "model": "default", "transport": "subscription:anthropic-max" },
"frontier-metered": { "adapter": "claude-code", "model": "default", "transport": "api:anthropic-batch" }
```

Same adapter, same model, different transport. Today both invoke the identical `claude -p …` command with the same `HOME` and the same run-wide `modelAuthKey`, so:

1. **Cost routing is a fiction.** An operator who routes `maintain.calibration` to `frontier-metered` expecting batch-API pricing may silently consume the subscription instead (or vice versa). Nothing surfaces the discrepancy.
2. **The §7 fallback edge is inert.** `frontier → frontier-metered` exists *only* to change transport. Validation rule 3 enforces that the two transports differ — which currently guarantees nothing at execution.
3. **Records can be untrue.** The dry-run prints `transport=…`, and §6's attestation schema is specified to record it. A recorded transport that execution never honoured is exactly the "label proves nothing about what ran" failure §4c was written to close for `model`.

Note the asymmetry: §4c already solved this for `model` (force it onto the command line, attest what actually resolved). `transport` has the force half missing entirely.

## Scope note

This is **not** a regression — transport has never been load-bearing. T151 built the validation and routing halves; the consumers of `transport` are later build items:

- §6 quorum gateway-discounting → spec §9 item 4 (attestation schema v2)
- §7 fallback traversal → not yet ticketed

What T151 changed is that transport is now *displayed and recorded*, which makes the gap visible and worth closing.

## Proposed fix

Extend the WorkerAdapter contract so transport selects a concrete execution context, and pass the trusted seat transport into dispatch:

[...]

PR #444, opened 2026-08-02
fix(fleet): re-derive the dispatch seat per attempt so a failed strike climbs the ladder (#401)

### LINKED ISSUE #401: Ladder escalation is inert: a strike-2 retry re-dispatches on the same channel
## Summary

Spec `operating-stack.md` §5 describes `build.ladder-start` as routing to the assignment tier's channel, with **"Escalation per F8 is a fresh attempt"**. Nothing implements that escalation. A ticket that fails its gate is re-dispatched by the two-strike policy onto the **same** channel and the same model, so the ladder never climbs.

Surfaced by an adversarial-review round on the T151 branch (build item 2 of spec §9).

## Scope note — not a regression

Fleet has never escalated models between strikes: `packages/fleet/lib/scheduler.mjs` tracks `strikes` but contains no tier or model logic, and `run.mjs` re-dispatches with the same deps. T151 added per-ticket channel resolution but resolves each seat **once per run**, so it inherits the existing behaviour rather than changing it.

The gap is newly *visible* because the registry now makes "which model ran this attempt" an explicit, reported property.

## Why it matters

The ladder's whole economic premise is that slack work starts cheap and escalates on failure — that is what makes a cheap first attempt rational. Without escalation:

- a ticket that a cheap model cannot complete burns **both** strikes on that model and fails, where the design intends the second attempt to be a stronger one;
- `assign.mjs`'s `ladder` budget already provisions for this (`resolveBudget` returns *starting tier + one frontier regeneration*), so the budget is allocated for an escalation that never happens;
- F8 ("escalation is a fresh attempt, never a rescue") is stated as a design property in several places but is unenforced.

## Proposed fix

Re-derive the seat per **attempt**, not per run:

- On a strike-2 dispatch for a `build.ladder-start` ticket, route the next channel up (`cheap` → `mid` → `frontier`) and dispatch there.
- Keep it a *fresh attempt* per F8 — a new worktree state and the fix charter, not a continuation.
- Emit an event so the escalation is visible (spec §11 q4 explicitly leaves the shape open: likely `kind: "escalation"`, distinct from a §7 fallback traversal).
- Leave `direct`-mode tickets alone; only the ladder escalates.

## Acceptance criteria

1. A ladder-start ticket that fails strike 1 dispatches strike 2 on the next channel up, asserted on the argv actually rendered.
2. A `direct`-mode ticket re-dispatches on the same channel.
3. Escalation stops at `frontier` rather than running off the end of the ladder.
4. The dry-run states the starting channel and that escalation may occur, so the plan does not imply a single fixed model.
5. Spec §11 open question 4 resolved and §5's escalation row updated to point at the implementation.

## References

- Spec: `docs/specs/operating-stack.md` §5 (routing table), §11 q4 (event shape), F8
- Code: `packages/fleet/lib/scheduler.mjs`, `packages/fleet/lib/run.mjs`, `packages/fleet/lib/quartermaster.mjs` (`planSeats`), `packages/model-router/lib/assign.mjs` (`resolveBudget`)
- Ticket: T151

voodootikigod:
Closes #401. Ticket `T-01KZ1ZR79FVPTBD847C6RTBSG8`.

**Depends on #443** — the §5/§11 spec edits below are inside `docs/specs/operating-stack.md`, which T151's rails froze for every PR in the repo. `rails-guard` on this branch stays red until #443 lands.

## The gap

Spec §5 routes `build.ladder-start` to the assignment tier's channel and states *"Escalation per F8 is a fresh attempt."* Nothing implemented it. `planSeats` resolved each ticket's seat **once per run**, and `advanceTicket` re-dispatched with the same deps — so a ticket a cheap model could not finish burned **both** strikes on that cheap model and failed.

That is the ladder's entire economic premise. Slack work starts cheap *because* a failure escalates; without escalation a cheap first attempt is only a cheaper way to fail. `assign.mjs`'s `resolveBudget` already provisions for it (`ladder` mode returns starting tier **plus one frontier regeneration**) — budget allocated for an attempt that never happened.

## What changed

**`@adlc/quartermaster/lib/escalate.mjs`** (new) — the pure §5/F8 rule. Ladder is `cheap → mid → frontier`; attempt 1 is the routed channel, each later attempt climbs one rung and **stops at the top** rather than indexing past it. `rungForAttempt` is generic over the rung payload, so the channel-level rule and fleet's seat-level resolution are literally the same index computation and cannot drift.

**`planSeats`** resolves every rung against the registry at planning time — dispatch needs no registry, and the dry-run can show the whole ladder.

**`live-deps`** dispatch selects the rung for the strike in hand. The transcript header now names the channel, because with escalation two strikes of one ticket can be two different harnesses.

**Both `job` and `mode` gate escalation.** A ticket below the rail-density floor is assigned `{tier: 'frontier', mode: 'direct'}` while `routeJob` still *derives* `build.ladder-start` from its non-zero float. Keying on the job alone would escalate a ticket the router deliberately routed direct. Asserted on the rendered argv, not on a plan object.

## Two defects this surfaced — fixed here, not shipped

**1. Escalation would have run an unprovisioned harness.** `provision()` runs once, before the pipeline, and provisioned only the *starting* rung. A ladder that crosses adapters (cheap/mid on opencode, frontier on claude-code) would have put the escalated strike on a harness whose `provision` never wrote the worktree allowlist its worker needs — indistinguishable from a model failure at the point it bites. Every distinct adapter on the ladder is now provisioned.

**2. The ledger would have lied.** The §8a dispatch carrier recorded `entry.route.channel` — the **starting** channel — for every strike. An escalated strike would have been recorded, and priced, against the cheap seat's channel *and transport*. That is exactly the "the label proves nothing about what ran" failure §4c closed for `model`, and the one #396 describes for transport. The carrier now records the channel that ran plus an `escalatedFrom` sibling, present **only** when the strike actually climbed — so the key's absence is itself evidence.

This required threading the strike number through `scheduler → buildEffects → recordDispatchUsage`, covered by a `runFleet`-level test rather than a direct call, so the wiring itself is what is asserted.

**Hardening on that new seam:** an unusable strike number resolves to *no seat* instead of throwing. `recordDispatchUsage` runs inside a best-effort `try/catch`, so a throw would have discarded the whole carrier and lost a real call's spend in silence — the same "unknown silently becoming zero" this recorder already warns about for CLI failures. The counters survive; only the provenance drops, and it says so on stderr.

## Spec

- **§5a added** — the escalation contract (ladder order, stop-at-top, the `job`+`mode` rule, provisioning, and the carrier requirement).
- **§5's routing row** points at it.
- **§11 open question 4 RESOLVED**: no separate `kind: "escalation"` event. The §8a carrier already exists, is written exactly once per dispatch, and already carries `channel`/`transport`/`registryDigest`. A second producer for the same fact would invite the double-counting the one-carrier-per-call rule exists to prevent, and no §8c monitor consumes such an event — §8b's `fallback` covers §7 traversals, and `seatRole: "escalation"` is a reviewer-seat property, not this.

`packages/model-router/lib/assign.mjs` is unchanged: its `{tier, mode, budget, float}` output is consumed exactly as emitted.

## Verification

- **29 new tests**, each verified RED with the fix reverted (not assumed — reverted and re-run).
  - `packages/quartermaster/test/escalate.test.mjs` — the pure rule, including a pin that the ladder holds exactly the channels `routeJob` can start a ladder on.
  - `packages/fleet/test/ladder-escalation.test.mjs` — seats built by the **real** `planSeats` over real registry bytes (real `computeFloat`, real `assignAll`, real `routeJob`); every dispatch assertion is on the argv the adapter actually rendered.
  - `packages/quartermaster/test/escalation-dryrun.test.mjs` — the real `fleet run --dry-run` CLI in a throwaway repo, human and `--json` output.
- **`mutation-gate`: 7/7 killed, 0 survived.** The first run had a survivor — `seat.escalation?.length > 0` → `> 1` — which was a *genuine* coverage gap: no test covered a **single-rung** ladder, i.e. the ordinary `mid`-start case, where escalation would have been hidden from the plan. Fixed with a real fixture, then confirmed by hand-applying the mutant and watching the new test fail.
- **`node scripts/run-tests.mjs`** → exit 0.
- **Adversarial review, 2 rounds, `agy` (Gemini 3.6 Flash High — distinct provider from the `anthropic` author): APPROVE, zero findings** both times. Round 2 was focused on the specific risk surfaces (recorder default, index-math boundaries, provision loop partial failure, the new `live-deps → quartermaster` import, whether escalation can reach an unauthorized channel or `frontier-metered`, and whether the new tests are hollow).

**Reviewer availability note:** `copilot` is over its monthly quota and `codex` resets Aug 8, so a second *family* was not available. Both rounds are the same provider family; the "diversity beats repetition" lesson argues for another family before this is treated as strongly reviewed.

## Not in scope

**#396** — `transport` is recorded but still not load-bearing at dispatch, so two rungs differing only by transport execute identically. The carrier now records the escalated rung's transport truthfully, which makes that gap *more* visible, not less.


voodootikigod:
## Both red checks are expected, and both clear the same way

| Check | State | Why |
|---|---|---|
| `test` (18/20/22), `mutation-gate`, `ticket-store-platform` ×9, `codex-live-latest`, `opencode-live-latest`, `router-drift`, `pre-ga-gate` | ✅ | — |
| `rails-guard` | ❌ | `[rail-edit] docs/specs/operating-stack.md` — frozen by T151's rails. Clears when **#443** merges. |
| `gate` | ❌ | Genuine TRUST-ROOT (real `packages/**` changes) with no attestation recorded **yet** — deliberate, see below. |

### Why the attestation is not recorded yet

The review has already happened: **two `agy` rounds (Gemini 3.6 Flash High — distinct provider from the `anthropic` author), APPROVE with zero findings**, the second one focused on this change's specific risk surfaces.

But the attestation binds to `git-change:<baseSha>:<contentDigest>`, and merging #443 changes this branch's base. Recording now would produce an entry that the required rebase immediately voids — and then a second entry for identical content. So the order is:

1. #443 merges (needs an admin override — `rails-guard` refuses a completion annotation on a railed ticket by design).
2. Rebase this branch onto the new `main`.
3. Confirm the change-set **content digest is byte-identical** — only the base SHA differing — which is the documented condition for the existing SHIP verdict still applying.
4. Record the attestation against the new revision, with `--ticket T-01KZ1ZR79FVPTBD847C6RTBSG8`.

### Reviewer-diversity caveat, stated plainly

`copilot` is over its monthly quota and `codex` does not reset until **Aug 8**, so both rounds came from the *same* provider family. The repo's own hard-won lesson is that diversity beats repetition — five rounds against one model went quiet where the first round against another family immediately found a HIGH. Two same-family rounds should be read as weaker assurance than "two independent reviews," and a codex or copilot round before merge would be worth the wait if it is available.

The deterministic evidence is stronger than the model evidence here: `mutation-gate` 7/7 killed (after a **real** survivor exposed a missing single-rung-ladder fixture), 53/53 suite segments, and every one of the 29 new tests verified red with the fix reverted.

PR #466, opened 2026-08-08
chore(adlc): ticket — gate-manifest migrate-branch (in-flight branch salvage after cutover)

voodootikigod:
Ticket document only, no code. Adds `T-01KZHAG1NSKDQTRB1JHZ4EJY2M`, edged as a prerequisite of `T-MANIFEST-FOREST-MIGRATE`.

## The scenario it covers

Main cuts over to the segmented manifest while branch-a is still developing. branch-a rebases and hits a **guaranteed** conflict at the root tail (reproduced in a scratch repo — seals+cutover vs the branch's entries, both appended after the same ancestor line). The only correct resolution is taking main's side wholesale: a union breaks the hash chain, fails today's `assertAppendOnly`, and fails §9.1's byte-identical rule once slice 6 lands.

The spec's answer (§8 step 7) is drop-and-re-record. That silently breaks `prosecute record-cross-model --carry-forward`: the prior approve is no longer findable in the manifest, so a clean rebase that changed no diff content still costs a fresh distinct-provider review.

## What the ticket specifies

`gate-manifest migrate-branch` — key-required, dry-run by default: verify the working root was resolved to main's side; read the pre-rebase root from `ORIG_HEAD` (or `--from`); signature-verify the branch-unique suffix; re-chain it into a fresh segment anchored to the cutover line (content fields byte-preserved, `seq`/`prev`/`sig` new); disclose the re-signing with a repair-chain-style salvage entry. Output satisfies §9.1/§9.3 as already specified — no rails-guard changes.

## Verification

- `adlc ticket create` dry-run: `validation: valid`, create-only, no graph effects
- `adlc coldstart --prompt-only` audit answered: no gaps (the simulation, preconditions, reuse targets, and per-AC commands are embedded; the spec is in-repo)


voodootikigod:
Superseded: the ticket shard this PR adds shipped inside #475 (cherry-picked with the implementation, byte-identical shard on main). Nothing here remains to merge — the diff vs main is now pure deletion of #475's work.

PR #477, opened 2026-08-10
feat(context-handoff): enforcing-tier adapters — Codex, OpenCode, Pi (slice 5)

voodootikigod:
Implements **slice 5** of `docs/specs/context-rot-handoff.md`: the remaining
**enforcing tier** — Codex, OpenCode and Pi — plus the shared adapter core they
import instead of forking.

Ticket: `T-01KZNZE8SQFBJ6S4C6W8QFZFP2`. Slice 4 (CC) shipped in #476.
Advisory/manual tiers (Cursor, Gemini/Antigravity, Copilot) are slice 6; Proven
CI is slice 7.

## What lands

**1. Shared adapter core → `@adlc/context-handoff`.**
`evaluateHandoffPreToolUse`, `isProtectedHandoffPath`, `isHandoffMutatingShell`,
`resolveHandoffSessionId`, `denyStoreHot` and `HANDOFF_MUTATING_SUBCOMMANDS`
move out of the Claude Code plugin into `lib/adapter.mjs`. All four harnesses
now evaluate D1–D3 through one implementation.

The CC hook module keeps behavioural twins of two pure helpers, because the
frozen slice-4 contract test drives them synchronously and that module cannot
resolve the package synchronously (no workspace `node_modules` in a plugin
install dir; Node 18 cannot `require()` an ESM package). They are pinned by
`adapter-test/cc-helper-drift.test.mjs`; the hook's real decisions run through
the package. The frozen test passes **unmodified**.

New package tests live in `packages/context-handoff/adapter-test/` so they do
not touch the frozen `test/**` rail glob.

**2. Codex** — `hooks/adlc-handoff-gate.mjs` on the existing PreToolUse mutation
matcher. Hook scripts cannot bare-import an npm package from their install dir,
so `handoff-resolve.mjs` is copied and pinned byte-for-byte by a drift test (the
pattern `packages/core/test/shell.test.mjs` already uses). Session identity:
payload `session_id`, else the `transcript_path` stem.

**3. OpenCode** — wired into `tool.execute.before` **ahead of** the rail check,
keyed on the host's `sessionID`, with depth from the in-process tracker the
build-gate backstop already maintains. Also honours the dormant `permission.ask`
lever.

**4. Pi** — wired into `tool_call` before every other gate. Pi is the first
adapter with a **real** context-fill signal: `ctx.getContextUsage().percent`
feeds the band join as `pct` rather than a transcript-size proxy. Session
identity is host-supplied when offered, otherwise minted per extension instance.

## Three decisions worth reviewing

**The deny-set is not ticket-scoped (Pi).** Every other Pi gate returns early
without an active ticket. An open deny record is a fact about the *session*, so
the handoff check runs before that early return. Pinned by a test.

**`advisoryHooks` does not downgrade the handoff deny (OpenCode).** That escape
hatch weakens the *rail* guard, which has an authoritative CI backstop. The
deny-set has none — its only exits are a signed `adlc handoff resume`, a signed
bypass, or privileged host repair. An env var must not be a fourth one. Pinned
by a test.

**Pi's build-gate percent deny is now shadowed.** Pi's build-gate threshold (80)
sits above the handoff band (60), so any percent that would trip the build gate
trips the deny-set first — including with `ADLC_BUILD_GATE_BYPASS=1`, which does
not clear a deny-set. Rather than rewrite the four affected tests to match, their
assertions were moved onto paths that are still reachable (`checkBuildGate`
directly, and the compaction signal, which degrades a session without moving the
percent), and a new test pins the precedence explicitly.

## Two gate findings, fixed rather than suppressed

- **mutation-gate, survived:** the Codex hook's `MAX_SCAN_BYTES` equalled
  `HARD_BYTES` by construction — a transcript large enough to trigger windowing
  already denies on the bytes signal, so no behavioural test can distinguish
  256KiB from 257KiB. The hook now takes the window from the loaded package,
  which removes the literal, the mutant, and the threshold-scan allowlist entry.
- **ticket-store writer boundary:** the guard skipped `test/` but not sibling
  trees, so `adapter-test/` was read as production code and flagged for naming
  `.adlc/tickets.json` in a fixture. Widened to every `*-test` directory, with
  both directions pinned.

## Verification

```
npm run preflight     # 6/6 PR-blocking gates, 54/54 test segments, 9/9 mutants killed
```

Per-criterion:

```
node --test packages/context-handoff/adapter-test/*.test.mjs          # 31
node --test plugins/adlc-codex/hooks/test/handoff-deny.test.mjs       # 18
node --test plugins/adlc-codex/hooks/test/handoff-resolve-drift.test.mjs
node --test plugins/adlc-opencode/test/handoff-deny.test.mjs          # 19
node --test plugins/adlc-pi/test/handoff-deny.test.mjs                # 23
node --test plugins/adlc-claude-code/hooks/test/handoff-deny.test.mjs # 23, unmodified
adlc spec-lint docs/specs/context-rot-handoff.md                      # 8/8 verified
git diff --quiet origin/main -- docs/specs/context-rot-handoff.md .adlc/specs/context-rot-handoff.md
```

Binding specs untouched. `tier-check` reports **trust-root** (rails deny-paths
of T37/T38/T53), so a distinct-provider cross-model approve is required before
merge.


voodootikigod:
## Cross-model review: three rounds, six defects fixed, one open

`npx adversarial-review --base origin/main --provider codex` was run three times.
Every finding was verified against the source rather than accepted or dismissed
on the reviewer's own confidence label.

**Fixed (all verified real):**

| Round | Severity | Finding | Whose |
| --- | --- | --- | --- |
| 1 | HIGH | Pi minted a session id instead of using `ctx.sessionManager.getSessionId()`; a reload dropped D2 | this slice |
| 1 | HIGH | `authorized()` needs `resumeAuth.verified === true`, but no adapter passed a key — a completed `adlc handoff resume` could never clear its own deny | pre-existing (#476) |
| 1 | MEDIUM | `processStickyDeny` is a per-call local: a failed marker write denied one call, then failed open | pre-existing (#476) |
| 2 | CRITICAL | the round-1 key fix handed `ADLC_MANIFEST_KEY` to a module resolved from project space | introduced by the fix, reverted |
| 2 | HIGH | `import.meta.url === \`file://${argv[1]}\`` — the hook silently no-ops (exit 0 = allow) from any install path with a space, and on Windows | this slice |
| 2 | MEDIUM | widening the ticket-store writer boundary to `*-test` took the shipped `packages/hollow-test` out of scope | this slice |
| 3 | MEDIUM | the handoff hook shipped a second `countToolCalls` that missed the prose tool-log form the canonical Codex counter recognizes | this slice |

**Rejected, with reasons:**

- *"A missing transcript is treated as a healthy signal"* (raised twice). Deliberate:
  `plugins/adlc-claude-code/hooks/test/handoff-deny.test.mjs` asserts
  "missing transcript_path file does not invent hard-band deny", and the spec's
  join ignores absent signal kinds so a dead telemetry source cannot hard-lock
  the repo. That test is a **frozen rail** on this PR.
- *"A recorded bypass has no channel into the gate"*. By design — `handoff bypass`
  documents its grant as "for the calling adapter invocation only; it is not a
  stored credential", so a hook has nothing to discover.

**Open — needs a maintainer decision (CRITICAL, pre-dates this slice):**

`handoff-resolve.mjs` resolves `@adlc/context-handoff` from the **project's**
`node_modules` and imports it into the hook process. A repository shipping a
package under that name gets code execution inside an enforcing hook, and that
process inherits the environment — including `ADLC_MANIFEST_KEY` when the
operator has one exported. Passing `manifestKey: null` (round 2) closes the
explicit hand-off but not the inherited environment.

This is the slice-4 resolver design, shipped in #476 and used by the Claude Code
hook today; this PR adds a second consumer. A real fix means bundling the
enforcing core at a plugin-owned immutable path and/or scrubbing secrets from
the hook subprocess environment — an architectural change to how the plugins
ship, affecting Claude Code too. It is out of this ticket's declared scope, so
it is flagged here rather than absorbed silently.

**No attestation has been recorded.** The third round's gate output derives
`APPROVE` from halved confidences while the model verdict is `needs-attention`;
recording on that basis is exactly the failure mode the repo has hit before. The
trust-root gate stays red until the CRITICAL above is adjudicated.

voodootikigod:
## Cross-model review: five rounds, what they found, and why no attestation is recorded

`npx adversarial-review --base origin/main --provider codex --timeout 2400`, run to convergence. Posting the round-by-round record so the residual is visible without re-deriving it.

**The gate's mechanical verdict is not the model's verdict.** It derives `APPROVE` from halved confidences while the model verdict is `needs-attention`. That is the confidence-halving trap this repo has been burned by before, so nothing was recorded off it. Every finding below was verified against source before it was touched.

### Round 1 — `8661bb3b`

- **Pi minted a session id per extension instance.** `ctx.sessionManager.getSessionId()` exists on the `ReadonlySessionManager` every `ExtensionContext` carries; the resolver skipped it, so an extension reload handed the same rotten session a fresh identity and dropped D2. The managed id is now the primary candidate; the mint survives only for contexts exposing no session manager, and still fails closed when unusable.
- **Three integration pages overclaimed shell protection.** They said deny-store/auth/stamp/lock writes are refused whether or not a deny is active. True for structured writes only.

### Round 2 — `ec9431b5`, `6fa3ec95`

- **A signed resume could never resume.** `authorized()` requires `resumeAuth.verified === true`, but every adapter called `readResumeAuth` without a key, so `verified` was always false — a completed `adlc handoff resume` could not clear the deny it was minted to clear. Adapters now pass their key; forged signatures and stale content hashes stay denied.
- **D1 was not sticky across a failed marker write.** `processStickyDeny` was a per-call local, so a marker write that failed denied one call and then failed open once the band cooled. `denyEverWritten` is now caller-threaded.
- **CRITICAL — threading `ADLC_MANIFEST_KEY` into the hook handed the signing key to project-controlled code.** A regression this PR introduced and reverted: `handoff-resolve.mjs` resolves the package from the *project's* `node_modules` by design, so a repo shipping `@adlc/context-handoff` could have exfiltrated the manifest trust anchor.
- **HIGH — the direct-execution guard silently disabled the hook.** It compared `import.meta.url` against a string-concatenated `file://` + `argv[1]`, so on any install path containing a space — and every Windows path — `main()` never ran and the hook exited 0, read as ALLOW. Now `pathToFileURL`, with a subprocess test from a spaced directory that fails against the old form.
- **MEDIUM — a `*-test` skip took a shipped package out of the writer-boundary scan.** Widening the ticket-store writer boundary to every `*-test` directory removed `packages/hollow-test` from scope entirely. Back to an explicit set of suite directory names, with an assertion that `packages/hollow-test/lib` is still scanned.

### Round 3 — `f1b415a6`, `3e20dd53`, `94e7deea`

- **A duplicate `countToolCalls` disagreed with the canonical one.** The handoff hook shipped its own regex, missing the `Writing|Editing|Created` prose tool-log form the Codex build-gate counter recognizes: a session in that transcript shape read as depth 0 to the handoff band while the build gate saw it as deep. It now imports the canonical counter, with a drift test pinning that it never re-declares one.
- **A comment claimed the deny-store sentinel covers the durable half of D1.** It only exists once a marker write *succeeds*, so it covers nothing in the case the note was about.
- **The exfiltration half of the CRITICAL, closed.** Both enforcing hooks now delete `ADLC_MANIFEST_KEY` and `ADLC_ADMIN_KEY` before the import — proven by planting a hostile `@adlc/context-handoff` and reading what it saw. The fixture runs the hook from a plugin directory with **no `node_modules` ancestor**, the deployed layout, because inside a repo checkout the plugin's own walk-up finds the real package first and the fixture would prove nothing. A companion test asserts the planted package *is* still imported, documenting the bypass half as open rather than implying it is closed.
- **A shell could erase the deny store.** Shell was gated only when D1–D3 already denied the caller, so a session whose record had just been consumed could `rm -rf` the handoff tree and the deny-store sentinel. The docs claimed CI caught that; it cannot — those paths are gitignored, so the deletion never reaches a diff. The gate now refuses a shell command naming a protected path regardless of deny state, including the directory-level form.
- **Pi's `safeUsage` collapsed a throwing `getContextUsage()` into the same `null` an absent one produces**, which the band join reads as no-signal and lets through.

### Round 4 — `c426dacf`

- **HIGH — symlink aliasing bypassed the path guard.** `isProtectedHandoffPath` is lexical and all three adapters relativize with `resolve()`/`relative()`, so a repo could point an innocuous name at the deny store and have the gate approve the alias while the filesystem followed the link. Rails-guard resolves both forms for exactly this reason. Paths are now classified by their own name **and** by what they resolve to, walking up to the nearest existing ancestor so a not-yet-created file inside a symlinked directory is caught. Denials say which form matched.
- **The OpenCode page listed a signed bypass among the ways out of a deny-set.** That grant is scoped to the calling invocation and is not a stored credential, so an automatic hook cannot honour it.

### Round 5 — `f4821348`

- **A supplied-but-unreachable transcript failed open.** `observeHandoffSignals` treated a `transcript_path` that was supplied but did not exist the same as no `transcript_path` at all — an empty observation the band join reads as no pressure, so a rotated, deleted, or permission-denied transcript let a session well past the band mutate. Absent still means absent; supplied-but-unreachable is now an invalid signal, which fails closed.

## What survives — two architectural gaps, both now tracked

Neither is introduced by this PR; both are inherited designs this slice extends to more harnesses.

**Gap 1 — enforcing hooks execute project-resolved code (CRITICAL).** `handoff-resolve.mjs` resolves `@adlc/context-handoff` from the project's `node_modules` and imports it inside an enforcing hook. The exfiltration half is closed (round 3). The **bypass** half is open: a repo shipping a package under that name can export the checked shape and return `deny:false`. Other ambient credentials also remain visible. Tracked by `T-01KZPRW2TBKCV66AZ00KYV6603` (#481), whose criteria demand a planted-package test proving the hostile module is *never imported*.

**Gap 2 — deny state lives in agent-writable project storage.** The round-3 fix narrows the shell path but does not close it: the scanner reads literal tokens, so one shell-variable indirection defeats it. Same root cause as the hook subprocesses hard-coding `denyEverWritten: false`. Both need host-owned storage the agent's shell cannot reach. Now tracked by `T-01KZRCNX3TSJ4C0PXZ28C9CB5N` (#487) — it was untracked when this PR was last pushed.

## Attestation

Not recorded, and it should not be recorded against this revision — the attestation binds to `git-change:<base>:<digest>`, and this branch still has to rebase onto #487. The final review runs against the merge-ready revision, author provider `anthropic`, reviewer distinct (`codex`, `agy`, or `copilot`).

Everything else is green: `npm run preflight` 6/6 gates, 54/54 segments, 10/10 mutants killed; CI green on every check except `gate`.

voodootikigod:
## Rounds 6–12: five more defect classes, and the basis for the recorded attestation

Continuing the [round 1–5 record](https://github.com/voodootikigod/adlc/pull/477#issuecomment-5253178159). After rebasing onto #487 the review was re-run to convergence against each successive revision. Every finding below was verified against source — by probing the real module, not by reading the report — before anything was changed, and each fix was confirmed to bite by reverting it and watching the new test fail.

### Round 6 — `ff458c3a` · shell path protection decided on the spelling, not the path

The directory-coverage expansion ran on the **literal command token** while protection was decided on the **resolved path**, so only the exact strings `.adlc` and `.adlc/handoffs` were ever expanded to their protected children:

```
rm -rf .adlc/handoffs              denied
rm -rf .adlc                       ALLOWED   (slashless token dropped entirely)
rm -rf ./.adlc                     ALLOWED
rm -rf /abs/repo/.adlc             ALLOWED
rm -rf /abs/repo/.adlc/handoffs    ALLOWED   (the denied command, respelled)
```

These are literal paths with no variable and no expansion — precisely what the guard claimed to cover, so not the documented best-effort limitation. `classifyShellTarget` now answers on the resolved repo-relative form and asks whether the target *reaches* a protected artifact.

### Round 7 — `39c6e13d` · the round-6 fix denied ordinary commands

A regression introduced by the previous round. Treating the repo root as covering every protected artifact was wrong, because candidates include every **ancestor** of every token: `.` is synthesized by any dot-relative path and the repo root by any absolute in-repo path. `cat ./src/app.mjs`, `node ./scripts/check.mjs`, `git -C . status`, and any absolute source path were denied on a clean repo, across all four adapters.

The round-6 suite missed it because every ordinary-command case used a bare relative path (`cat src/app.mjs`), which synthesizes no ancestor at all. Those cases now cover dot-relative and absolute spellings in both directions.

A command that erases the whole checkout is no longer treated as reaching the deny store — it cannot be told apart from the synthesized ancestor of an ordinary path, and denying the latter to catch the former is the worse trade.

### Round 8 — `08e5d14c` · parallel tool envelopes escaped the shell scan

`hooks.json` routes `multi_tool_use.parallel` through the handoff hook, but the hook decided `isBash` from the **outer** tool name only. A parallel envelope is not itself a shell tool, so a nested `exec_command` reached the core with `isBash:false` and an empty command string — skipping both the always-on path scan and the wholesale shell block.

```
direct   exec_command  "rm -rf .adlc/handoffs"  → isBash=true,  scanned, denied
parallel envelope, same command nested          → isBash=false, cmd="", allowed
```

`hasShellToolAnywhere` walks the payload for a nested shell recipient — the shape `adlc-rails-guard.mjs` in this same plugin already uses. Edit-path collection is now skipped only for a *direct* shell tool, so an envelope carrying both a nested `apply_patch` and a nested `exec` gets both checks.

### Round 9 — `ecad67d2` · pi vetted handoff targets with a narrower extractor than its own rail gate

pi's rail gate vets custom tools through `extractToolPaths`, which reads `target` and `file` as well as the three `path` spellings. The handoff gate read only `path`/`filePath`/`file_path`, so a third-party tool reached the deny store with an empty `editRelPaths` while the rail checker saw the path perfectly well.

```
{path:   '.adlc/.deny-store'}             denied
{target: '.adlc/.deny-store'}             ALLOWED
{file:   '.adlc/handoffs/denies/d.json'}  ALLOWED
```

### Round 10 — `f3ed295c` · the same, in OpenCode

Round 9's write-up claimed OpenCode did not share this omission. That was wrong, and worth recording as such: OpenCode's handoff gate *does* reuse the plugin's `extractTargets` — which is what had been checked — but that extractor is itself narrower than pi's. It covers the shapes OpenCode's own mutators use (`path`, `file`, `filePath`, `files[]`, `edits[]`, apply_patch bodies) and not `target`/`targetPath`.

Widened in the handoff gate rather than in `extractTargets`, because that extractor also drives per-ticket rail policy whose behaviour is frozen. **The rails-side gap is real and remains open** — worth a follow-up ticket; it is not this ticket's to change.

### Round 11 — `b350cfc8` · structured targets had no ancestor coverage

A consequence of the round-6 design decision. Ancestor coverage went only into the shell classifier, on the reasoning that a structured target is a file rather than a directory. That holds for a harness's own `write`/`edit` tools. It does not hold for the third-party tools OpenCode and pi deliberately route through the same path, and a delete/move tool can name a directory:

```
{target: '.adlc/handoffs'}   ALLOWED        rm -rf .adlc/handoffs   denied
{path:   '.adlc'}            ALLOWED        rm -rf .adlc            denied
```

The integration pages claim structured writes to these artifacts are refused whether or not a deny is active, so this was a doc overclaim too. `classifyShellTarget` is renamed `classifyProtectedTarget` and serves both branches — the question it answers was never shell-specific. `classifyHandoffPath` keeps its narrow artifact-identity semantics and its frozen contract. Siblings holding nothing protected (`.adlc/tickets` and below) stay editable, pinned alongside the deny cases.

### Round 12 — clean

Three findings, all three the tracked architectural gaps. No new class, and the first round since 5 with no defect in this PR's own code.

## What is being attested, and what is not

The recorded verdict is `approve`; the reviewer's own verdict is `NEEDS ATTENTION`. That gap is deliberate and worth stating plainly, because the two are not the same claim.

The review will **never** return approve while these remain open:

| residual finding | tracked by |
|---|---|
| CRITICAL — enforcing hooks import project-resolved code | `T-01KZPRW2TBKCV66AZ00KYV6603` (#481) |
| HIGH — variable indirection erases the deny store | `T-01KZRCNX3TSJ4C0PXZ28C9CB5N` (#487) |
| MEDIUM — subprocess adapters lose D1 after a failed marker write | `T-01KZRCNX3TSJ4C0PXZ28C9CB5N` (#487) |

All three are inherited designs this slice extends to more harnesses rather than defects it introduces, both tickets are merged to main with acceptance criteria demanding planted-package and host-owned-storage proofs, and the maintainer decision was to merge this slice with both tracked. The attestation records that judgment against `git-change:32d511cb…:e9ff3f2d…` — not a claim that the reviewer approved.

`allow-suppression: none` still holds; nothing here suppresses a gate.

## Verification

- `npm run preflight`: 6/6 PR-blocking gates, 10/10 mutants killed, on every revision above
- Frozen contract suites unmodified and passing throughout — the round-11 change was checked against them specifically, since it was the one most likely to break them
- Each new test confirmed to fail against the pre-fix implementation before being kept

PR #502, opened 2026-08-13
fix(opencode): the frozen rail set must outrank a caller's claim about a target (T-01KZY120SZP96QKEEXDKZEZJR5)

voodootikigod:
Follow-up to #490 (`T-01KZW2DQKN5C1BANMMTTMPK8TS`, merged as e87c933d). Ticket: `T-01KZY120SZP96QKEEXDKZEZJR5`.

#490 taught the ungated-tool spoof branch to see target-keyed arguments and gave it a file-vs-directory distinction. Cross-model review found three defects in that distinction after the merge window closed. They share a root cause, and it is the one #490's own handoff wrote down: **an existing classification reused to answer a subtly different question.** Here the reused classification is the *caller's*. The ungated branch exists precisely because the tool name and argument shape are untrusted — and then it let that same shape decide whether ancestor detection applies.

Each was measured as ALLOW against e87c933d before it was fixed.

## 1. Aliased keys erased ambiguity

`spoofCandidates` deduplicates candidates by string but recorded provenance only for keys that asserted something, so a path arriving under both an ambiguous and a file-specific key kept only the file half. With a ticket railing `assets.bundle/**`:

```
{target: 'assets.bundle'}                            -> deny
{target: 'assets.bundle', filePath: 'assets.bundle'} -> ALLOW
```

Adding a spelling relaxed the check. Provenance now merges conservatively: any ambiguous occurrence revokes a file assertion about the same path, the way an explicit directory key already overrode one. Tested in both property orders.

## 2. A claimed file shed EVERY ancestor form

A file-keyed target went to `railHit` with `ancestors: false`. That is right for the form #490's requirement 3 targeted — `src/index.mjs` reading as an ancestor of `src/**/test/*.mjs`, an over-block created purely by the `**`. It is wrong where the rail set contradicts the caller: a rail `assets.bundle/**` *states* that `assets.bundle` is a directory, so `{filePath: 'assets.bundle'}` shed the check by asserting otherwise.

`railHit` gains `ancestors: 'literal'` — every ancestor form except the one an anchored `**` creates — and the ungated branch passes it for a claimed file. Two consecutive review rounds asked for opposite treatment of that key (trust it more, then trust it less); deciding it from the frozen rails rather than from the caller answers both.

## 3. Padded spellings were a different path

Extraction rejects a blank value with `trim()` but stored the raw string, and `railHit` canonicalized only that form. A padded ` test/frozen.test.mjs ` missed the rail while any tool that normalizes its input acts on the frozen one. **Not confined to the ungated branch** — with a ticket railing `test/**`:

```
task  {target:   ' test/frozen.test.mjs '} -> ALLOW
write {filePath: ' test/frozen.test.mjs '} -> ALLOW   <- the branch that writes files
```

Fixed inside `railHit`, where the ungated, mutating, shell, and gate-argv call sites all pass through.

## Deliberately not changed

`namesAFile`'s extension heuristic. A file-specific key *licenses* it and is not authoritative alone, because `filePath` is a key plenty of tools hand a directory. An absent `{targetFile: 'src/Makefile'}` therefore keeps ancestor detection and can be over-denied under an interior-wildcard rail. A review round asked for the opposite; the trade-off is now pinned by a test as a decision rather than left to be rediscovered — this predicate gates only tools that never write files, so a visible refusal is cheaper than a miss.

Also unchanged: the mutating branch's `singleFile` semantics (a known structured mutator writes the one file it names), the sibling pi/codex adapters, and the CI diff gate as the unbypassable backstop.

## Test plan

- [x] `node --test plugins/adlc-opencode/test/*.test.mjs` — 300 pass
- [x] `npm run preflight` — 6/6 gates, 12/12 mutants killed
- [x] `node scripts/mutation-gate.mjs origin/main --max 20` — 16/16 killed
- [x] Each defect probed as ALLOW against e87c933d, denied here
- [x] `plugins/adlc-opencode/test/handoff-deny.test.mjs` unmodified (railed by this ticket)
- [ ] Trust-root tier: distinct-provider cross-model approve — **pending**, must be recorded against the final revision


voodootikigod:
## Review rounds 7–15, and the one residual

Nine cross-model rounds (codex) ran against this branch. Every substantive finding was reproduced against a real revision before being fixed, and each fix is pinned by a test that was checked to fail against the revision before it. Summary of what they turned up, since the round-by-round commits are hard to read as a whole:

| # | Finding | Status |
|---|---|---|
| 7 | Aliased keys erased ambiguous provenance | fixed — conservative merge |
| 8 | A claimed file shed **every** ancestor form | fixed — `ancestors: 'literal'` |
| 8 | Padded spellings missed the rail, **including on the mutating branch** | fixed in `railHit` |
| 9 | Zero-width `**` hid the rail's own literal segments | fixed |
| 9 | Ticket prose read as instructions to a reviewer | fixed — descriptive wording |
| 10 | `**` absorbing a segment hid a literal naming | fixed — the question is the *last* segment |
| 10 | `foo-*` read as matching `bar` | fixed — same matcher as the direct check |
| 11 | Raw-vs-trimmed classification mismatch | fixed — one spelling list for both |
| 11 | Ancestor walk quadratic on a caller-chosen path | fixed — memoized |
| 12 | Still quadratic: a suffix scan per state | fixed — constant-work transitions |
| 13 | Deep rail overflowed the call stack | fixed — explicit worklist |
| 14 | Embedded `**` (`a/foo**bar/**`) lost its ancestor | fixed — regression vs merged main |
| 15 | Cartesian state space on rail × path | **residual, see below** |

Round 14's fix also exposed why enumeration was not converging: every one of rounds 10–14 was the same class — this predicate disagreeing with the matcher that decides direct hits. That is now a **property test** over generated rails rather than another remembered case: *if some path the rail matches has the target as a proper prefix, the target is an ancestor of that rail.* It failed on its first run and found two more misses (`a` and `a/a` under `a**b`) that no round had reported.

### The residual, and why it is tracked rather than fixed here

Round 15 reports a Cartesian blowup: the walk visits target-segments × rail-segments states, so a rail with hundreds of `**` segments plus a long caller-supplied path is expensive. Measured on the reviewer's own input (16,000-segment path × 200 globstars):

```
opencode ancestor walk   1029 ms
@adlc/core globMatch     did not finish in 10 minutes
```

`railHit` calls `globMatch` **first**, on this branch and on merged main alike. Bounding the walk therefore cannot change the outcome — the pre-existing matcher dominates by orders of magnitude, and it is reachable from every rail-enforcement path in the repo, not just this one. The blowup also needs a rail with hundreds of globstars, and rails come from the ticket store, which is a trust root: whoever can author that rail can delete the rail set instead.

Filed as **`T-01KZY8WXD86J1B2JJWARBZC9HD`** (core: globMatch backtracks catastrophically on a repeated-globstar rail) with the measurements, rather than papered over with a state budget in this adapter that would not help.

### Verification

- `node --test plugins/adlc-opencode/test/*.test.mjs` — 306 pass
- `npm run preflight` — 6/6 gates
- `node scripts/mutation-gate.mjs origin/main --max 24` — 24/24 mutants killed
- `plugins/adlc-opencode/test/handoff-deny.test.mjs` unmodified (railed by this ticket)

PR #515, opened 2026-08-15
fix(gate): P5 consumers read a causal chain, not forest array position (T-01M01HTKD290FQQB4NDCV60Q24)

voodootikigod:
Builds `T-01M01HTKD290FQQB4NDCV60Q24`.

> **Stacked on #514**, which files the ticket. Merge that first, or this diff carries the two ticket JSONs with it.

## The defect

`readManifestForest` orders segments by anchor topology — (anchored-to segment, anchored-to seq, own ULID) — so `show`/`attest` render the same forest the same way everywhere. Its own header says, in as many words, that this is deliberately **not** a causal or temporal order across independent segments. Three P5 consumers read array position as chronology anyway:

| where | how |
|---|---|
| `runner/lib/assertions.mjs:45`, `:277` | `latestP5Entry` / `latestP6Entry` → `.at(-1)` |
| `runner/lib/assertions.mjs:180`, `:199`, `:235` | `entries.slice(0, indexOf(entry))`, `slice(index + 1, completionIndex)` |
| `runner/lib/acceptance.mjs:20` | `latestP5Entry` → `.at(-1)` |
| `prosecute/lib/run.mjs:183` | sequential `Map.set` / `Map.delete` replay |

With one segment they are right by accident. With two, a foreign branch's ULID decides whether its entries land before or after ours. The dangerous direction is a **foreign kill sorting after our verification**: the replay deletes a finding nobody on this chain refuted, and three dry passes complete over the top of it.

## The fix

`readOwnManifestChain` (new, `packages/gate-manifest/lib/own-chain.mjs`) returns root's provably-prior prefix — bounded by the seq our segment's anchor names — plus this checkout's own open segment. That is the same root+own-segment scoping `cross-model.mjs`'s carry-forward already applies (`e.segment === 'root' || e.segment === recovered.name`), and that `tickets/lib/doctor.mjs` and `ticket-sync`'s reassign/push apply via `readOwnChains`. Own-segment identity comes from `recoverOpenSegment`, so it survives CI, where `.lineage` is gitignored.

It stays a **lenient** read like `readManifestForest` — trust is `verify()`'s job — but an **unidentifiable** checkout refuses on both channels rather than reading as "no evidence": `assertPhase` sees it in `skipped` (already `ok: … && skipped.length === 0`), and prosecute reads `identityError` and returns an op-error **before its first append**.

`readManifestForest` itself is untouched: its ordering is correct for what it documents, and display consumers keep it.

### Deliberately left on the whole-forest read

`recordedUsageByCall` asks "has this call already been recorded **anywhere**" — position-free, and missing a record would double-count real spend. Reading more is the safe direction there.

## Why a new file in gate-manifest rather than the three consumers

The ticket scoped the fix to the consumers. Triplicating this — segment identity, the anchor bound, the two refusal channels — across `runner` and `prosecute` would be three copies of security-relevant logic that must agree; `prosecute` cannot import from `runner` (no dependency), and both already depend on `gate-manifest`. `readManifestForest`'s own module is the only shared home. That is a deliberate widening of the declared scope, stated here rather than left to be noticed.

## Test plan

Every new case was run against the **unfixed** consumers to confirm it bites:

- `packages/runner/test/own-chain-chronology.test.mjs` — 4 of 6 fail unfixed. The 2 that pass are the controls: *"still reports OUR chain's own unresolved finding"* and *"reads evidence recorded in OUR OWN segment"*, which must behave identically either way.
- `packages/prosecute/test/own-chain-chronology.test.mjs` — 3 of 4 fail unfixed; the 4th is the control.
- `packages/gate-manifest/test/own-chain.test.mjs` — 9 unit cases for the reader itself, including "identical to the whole forest when the only segment is ours" and "identical in a repo that never segmented".

Suites:

- [x] `node --test packages/runner/test/*.test.mjs packages/prosecute/test/*.test.mjs` — 387 pass
- [x] `node --test packages/gate-manifest/test/*.test.mjs` — 439 pass
- [ ] `npm run preflight` — 6 gates
- [ ] CI green


voodootikigod:
## Review provenance, and its limits

Recording this because "cross-model approved" should not be read as more coverage than it had.

**Round 1 — codex (`9a56d1ab` vs `origin/main`): `needs-attention`, 2 findings, both real, both fixed in `8c4580e9`.**

The CLI printed `APPROVE`; that was the gate's *derived* verdict after halving both findings' confidence as ungrounded. The model's own verdict was `needs-attention`. Each finding was reproduced against the code before being acted on:

- **Detached HEAD (high).** `recoverOpenSegment` answers a detached checkout with `null` — the same value it uses for the benign "this branch has no segment yet" — so a `try/catch` cannot tell them apart, and a segmented ledger silently collapsed to root-only. A CI checkout of a PR SHA *is* detached, so this was live in exactly the environment the gates run in. `@adlc/tickets`' `readOwnChains` already refuses on the identical condition; this reader did not.
- **Segment-to-segment ancestry (medium).** The first version kept all of root plus the own segment and a comment called dropping an intermediate parent "conservative". It is not — keeping root's `p5-complete` while dropping a parent's still-open finding is a false pass. Conservatism only holds when evidence is dropped in both directions at once.

Fixing those surfaced two more of my own: `acceptance.mjs` read `entries` and ignored the refusal entirely, and a refusal returned root rather than nothing.

**Rounds 2–4 — the post-fix reads.**

| round | provider | outcome |
|---|---|---|
| 2 | codex | timeout at 900s, then 2400s — no verdict |
| 2 | copilot | CLI incompatible with this tool version (no stdin piping) — no verdict |
| 2 | agy (Gemini) | **approve**, 0 findings, whole diff |
| 3 | codex | timeout at 2400s (whole diff) and 1800s (fix commit alone, 30KB) — no verdict |
| 4 | agy (Gemini) | **approve**, 0 findings, focused on the rewritten ancestry walk |

**The limitation:** codex succeeded on a 57KB diff early in the session and then failed on a 30KB one, so it degraded mid-session rather than choking on size. This machine has no provider API keys, `copilot` is CLI-incompatible, and `claude` is disqualified as self-review — so **`agy` was the only working distinct provider for the post-fix reads.** The two findings above are the only strong-critic output on this branch, and they cover `9a56d1ab`, not the fix on top of it. The recorded attestation rests on Gemini.

A second independent provider re-reading `8c4580e9` would be worth having before this merges, if one can be made to run.

**Gates:** `npm run preflight` 6/6 green (54/54 segments; mutation gate 6 killed, 0 survived, all changed source files prosecuted). `tier-check` PASS against `git-change:40d8ad76…:15f5ad99…`.

PR #516, opened 2026-08-15
feat(tickets): doctor reports orphaned anchors and stale lineage tokens (T-01M01HTKJCNR8D8XTWR0GVC50K)

voodootikigod:
Builds `T-01M01HTKJCNR8D8XTWR0GVC50K`.

> **Stacked on #514**, which files the ticket. Merge that first, or this diff carries the two ticket JSONs with it.

## What was missing

`docs/specs/segmented-gate-manifest.md` §12 AC10 requires doctor to report orphaned anchors and stale lineage tokens, read-only. `doctorTicketStore` emitted `active-store`, `archive`, `current-ticket`, `storehash-manifest-bind`, `transactions`, `writer-lock` — and `orphan` appeared nowhere in `packages/tickets` or `packages/gate-manifest`.

Both faults are invisible to everything else **by design**. Every resolver swallows a broken `.lineage` token — `peekOpenSegment` falls through and the next writer mints a fresh segment, which is right for a writer and silent for an operator. An orphaned anchor surfaces only under a full `verify()`, which `adlc ticket doctor` does not run. Since the root was frozen, anchors and tokens are load-bearing on every branch, so the eventual failure lands far from its cause on any of them.

## The check

A new `manifest-forest` check reports:

- a segment whose anchor names a chain that is not in the forest, or a seq in that chain that no longer exists;
- a `.lineage` token naming a segment that is gone, or caching a ULID the segment's own name does not carry.

**Added** rather than folded into an existing check: names are read by operators and by ticket-store tooling. Inert (`segmented: false`) on a repo that never segmented.

It deliberately does **not** re-answer the integrity question. `MANIFEST_MALFORMED`, `MANIFEST_CHAIN_INVALID`, `MANIFEST_SIGNATURE_INVALID` and `SEGMENT_AMBIGUOUS` ask whether what *is* there verifies; this asks whether something it points at is gone. So an unparseable line contributes nothing here rather than becoming a second, weaker copy of a check that already exists — and `anchor: null` (a rootless segment, §4.4) is not a fault.

Read-only, like every other check in this file. An anchor repair path needs its own design and belongs to the migration and recovery ceremonies.

`discoverSegments` / `readLineageToken` / `ulidOf` are exported from `manifest-segments.mjs` so doctor applies the **same** derivations the resolvers do — a doctor with its own notion of which files count would report faults the writers never see and miss the ones they do. New exports need declarations, hence the `index.d.ts` hunk (`scripts/test/type-declarations.test.mjs` compiles a consumer importing the whole surface).

## Test plan

Seven cases in `packages/tickets/test/doctor.test.mjs`, all named `doctor manifest-forest: …` so the ticket's `--test-name-pattern='manifest'` reaches them. All seven fail against the unfixed `doctor.mjs`; the five fault cases fail on the assertion, not merely on the check's absence.

- healthy forest → `ok: true` (the check cannot pass by always firing)
- never-segmented repo → inert
- anchor whose root line was truncated away
- anchor naming a segment not in the forest
- `.lineage` naming a deleted segment
- `.lineage` caching a mismatched ULID
- **both faults at once → reported, and every byte under `.adlc/` is byte-identical after the run** (hashed before and after, whole tree, not a directory listing)

- [x] `node --test packages/tickets/test/*.test.mjs` — 349 pass
- [x] `node --test scripts/test/type-declarations.test.mjs` — 5 pass
- [x] `adlc ticket doctor` on this repo (14 segments) → `manifest-forest: ok`, no false fire
- [ ] `npm run preflight` — 6 gates
- [ ] CI green


voodootikigod:
## Review provenance, and its limits

| round | provider | outcome |
|---|---|---|
| 1 | codex | timeout at 1800s on a 37KB diff — no verdict |
| 2 | agy (Gemini) | **approve**, 0 findings — focused on false-negative/false-positive behaviour of the new check, `readChain`'s first-entry latch, read-only-ness, and the widened export surface |
| 3 | agy (Gemini) | **approve**, 0 findings — general whole-diff pass |

**The limitation, stated plainly: this branch has no codex read at all.** Codex succeeded on a 57KB diff earlier in the same session and then failed on smaller ones, so it degraded mid-session rather than choking on size. This machine has no provider API keys, `copilot` is CLI-incompatible with this tool version, and `claude` is disqualified as self-review — so `agy` was the only working distinct provider. Both approvals come from the same model.

That is weaker evidence than it looks: on the sibling PR (#515) codex found two real defects on a change Gemini later approved with zero findings, so Gemini is demonstrably the softer critic on this class of change. A second independent provider re-reading this diff would be worth having before merge.

**Gates:** `npm run preflight` 6/6 green (54/54 segments; mutation gate 6 killed, 0 survived). `tier-check` PASS against `git-change:40d8ad76…:15a70eb4…`.

**On the mutation gate:** the first run here surfaced a genuine survivor — `readChain`'s first-entry latch (`sawFirst = true` → `false`) was unverified because every fixture segment was one line long, making "first entry" and "last entry" the same object. The fixture now carries a second evidenced transaction, only the first entry holds the anchor (§4.4), and three tests fail if the latch is flipped.

PR #536, opened 2026-08-21
fix(tickets): doctor tolerates an honest legacy unsigned prefix, binds only to signed entries

voodootikigod:
`adlc ticket doctor` reported `store:storehash-manifest-bind — MANIFEST_SIGNATURE_INVALID` on this repository whenever `ADLC_MANIFEST_KEY` was exported. **The manifest is not corrupt.**

## Diagnosis

| | |
|---|---|
| Root entries | 271 — lines **1–95 unsigned** (2026-07-23 → 07-26), line 96 onward signed (from 2026-07-27) |
| Segments | 28, **fully signed**, zero bad entries |
| `gate-manifest verify --allow-legacy-unsigned` | `manifest ok (388 entries, 28 segments)` |
| `ticket doctor` with the key unset | All checks passed |

Those 95 entries are an honest legacy prefix predating HMAC signing. A missing `sig` is not a present-but-invalid one.

`walkChainForStoreHash` applied a blanket "with a key, every entry must be signed" rule and failed on line 1. That rule already exists correctly in **two** places — `gate-manifest/lib/verify.mjs` (`seenSignedEntry`, exposed as `--allow-legacy-unsigned`) and `chainIsIntact` in this package's own `manifest-segments.mjs`. `doctor.mjs` mirrors gate-manifest's `entrySigValid` byte-for-byte (it cannot import it: `tickets ← core ← gate-manifest` would cycle) but never mirrored the policy around it.

## The fix — two halves, deliberately paired

**Tolerate the honest prefix.** A *missing* signature is tolerated only before this chain's first valid signature. After that — and for any *present-but-invalid* signature at any position — behaviour is unchanged: hard failure. An attacker holding the tail of a signed chain still cannot regress it to unsigned by appending.

**Don't launder the tolerance into trust.** `storeHashBindingCheck` reports `authenticated: key !== null`. Making the walk lenient *without* this second half would turn that into a false claim — **trading a false-RED for a false-GREEN**, which is this suite's signature risk. So with a key set, only a **verified** entry may supply `boundStoreHash`. An all-legacy ledger now reports `bound: false` with a reason naming why, instead of the misleading "no evidence-required transaction recorded yet".

## Verification against the real ledger, not just fixtures

- `adlc ticket doctor` now passes with the key exported, and `bound: true, authenticated: true` is **truthful** — the checkpoint is root line 185, which is signed.
- **Tampered that signed entry in a copy of the actual manifest**: still `MANIFEST_SIGNATURE_INVALID`, forged hash never adopted. The leniency opened no hole.

The mutation gate first surfaced **5 survivors**, each a genuine gap — notably that a one-character `sig` was treated as *absent*, letting a forger opt back into legacy tolerance by truncating the field, and that `entry?.data &&` → `||` crashed on an entry with no `data` key. All five killed, **each verified by planting the mutant**.

| Gate | Result |
|---|---|
| doctor tests | 42/42 |
| Mutation gate | **20/20 killed, 0 survived** |
| Preflight (6 PR-blocking gates) | pass, 703s |
| adversarial-review (agy, distinct provider) | APPROVE, inline-diff, 4 files, no findings |
| tier-check | PASS |

## Deliberately not done

- **No `--allow-legacy-unsigned` flag on doctor.** The prefix rule is correct by default and strictly stronger than blanket leniency; a flag operators must remember is a worse default.
- **No re-signing of the 95 legacy entries** (`gate-manifest migrate --attest-unsigned`). That backdates trust in an audit ledger — asserting entries were signed when signing did not yet exist — and destroys the honest signal that they predate it.
- **No de-duplication of the three copies of this rule.** Real, but a refactor across a dependency cycle; out of scope.

## Note for the reviewer

Full suite is 53/54 on macOS, failing `context-handoff` › `continue --write captures…`. That is **pre-existing on `main`** (reproduces on a pristine `origin/main` checkout) — a macOS-only `/var` → `/private/var` realpath comparison in a test added by #532. Unrelated to this PR.

Ticket: `T-01M0JFM0TMM98GCNCN12BMD055`


voodootikigod:
Rebased onto `867b93cd` (main with #535 merged).

- Re-reviewed at the new base: adversarial-review `--provider agy`, inline-diff, 4 files examined, no material findings.
- Cross-model approve re-recorded and `tier-check` **PASS** at `git-change:867b93cd…:b40426be…`.
- Mutation gate: **20/20 killed, 0 survived**.
- `packages/tickets` on macOS (the `ticket-store-platform` job that gates this PR): **360/360**.
- The four manifest commits accumulated across the #532 / #534 / #535 rebase treadmill are squashed into one, with the erroneous `b6f811` entry disclosed in its message.

Full suite is 53/54 locally on macOS; the one red is `context-handoff` › `continue --write captures, binds, and consumes for exactly one successor`, which is **pre-existing on `main`** (a `/var` → `/private/var` realpath comparison in a test added by #532) and unrelated to this PR. `spec-lint` is green now that #535 has landed.

PR #538, opened 2026-08-22
chore(adlc): author the 1.11.0 release-blocker tickets (6 build units + 4 follow-ups)

voodootikigod:
## What

P0 output for the 1.11.0 release-unblock arc. `/release-audit` returned NO-GO (`release-blockers.md`: 8 blockers + the red spec-lint segment). Every blocker premise was re-verified live on `867b93c` (three parallel verification passes; the 3 post-audit commits touched only `plugins/adlc-claude-code` and `packages/context-handoff` — all 8 premises stand).

This PR authors the tickets only — no code changes. Builds follow as separate PRs per unit.

## Tickets

| Unit | Ticket | Blocker |
|---|---|---|
| U1 | T-01M0MX7YX0JJQS7M8S1SGG5BJS | rails-guard trust-roots breaking change (option (a) locked with Chris) |
| U2 | T-01M05BWTEYKHEK3JJX71NXXJ4H (refreshed, authorized scope-widening) | codex countToolCalls dead depth signal + byte-threshold raise |
| U3 | T-01M0MX7Z5Y5YFYYQAVWV1JPPSR | runner swallows own-chain identityError |
| U4 | T-01M0MX7ZEM8Y4QEDMGFHZYHJAY | opencode handoff lockout containment |
| U5 | T-01M0MX7ZQDP54AA2MJ0E1GCP7N | pi handoff lockout containment |
| U6 | T-01M0MX8002S0TQV3K0P18PF8GD | spec-lint doc-scan false positive (red suite) |
| U7 | T-01M0MX8092QVA41MX230AR0923 | build-gate README default + codex README CLI floor |

Follow-ups (post-release backlog): own-chain CI identity design (T-…7YB3), opencode depth epochs (T-…AJWYA), key-free recovery + recovery-exception port (T-…P65XX), shared-adapter contentHash:null (T-…NTY84).

## Gates

- Coldstart executability self-audit run per ticket (in-session, --prompt-only per the skill): the 7 release tickets are clean; F2/F3 carry deliberate design latitude as backlog follow-ups.
- t158 shard delta is store canonicalization only (escape → literal em-dash), byte-level no-op semantically.

https://claude.ai/code/session_011Fbv2GiKerPz67RapZigRM

PR #540, opened 2026-08-22
fix(opencode): handoff gate fires only in ADLC repos and writes a resumable, diagnosable deny

voodootikigod:
Ticket: **T-01M0MX7ZEM8Y4QEDMGFHZYHJAY** (release-blocker; ticket shards landed via #538).

## Blocker

Release-audit BLOCKER, verified by execution: in an empty directory with no `.adlc`, the OpenCode plugin's context-rot deny-set denied tool call 30 and everything after it — structured edits **and** the shell, repo-wide and cross-session — and it CREATED `.adlc/.deny-store` + `.adlc/handoffs/denies/<session>.json` in a repo that never opted into ADLC. The deny was also unrecoverable: the marker carried `ticket_id: null`, which `adlc handoff repair`/`resume` refuses outright, and the deny text named neither the session nor any command, while the agent shell it pointed at was itself inside the deny-set.

## What shipped

**Containment.** The deny-set now evaluates only in a repo that actually opted into ADLC, keyed on `ticketStoreExists` — this plugin's own definition of initialized, the same predicate `resolveRailsInForce` uses, so the deny-set and the rail guard no longer hold two different ideas of what an ADLC repo is. Deliberately *not* bare `existsSync('.adlc')` like the sibling adapters: a directory contaminated by the pre-fix bug has a `.adlc` holding nothing but the deny artifacts that bug wrote, so keying on the directory would let those artifacts vouch for themselves and leave exactly the bricked repos this fix exists to free still bricked — past reach of the agent shell needed to clean them.

An init latch keeps that signal honest. The ticket store is an unprotected file (`isProtectedHandoffPath('.adlc/tickets.json')` is false, and structured writes/deletes naming it are allowed), and the rail guard is inactive without `ADLC_P4_ENFORCEMENT` — so without a latch, containment hands the contained party an off switch. The latch can never arm a directory that never had a store, and never disarms one that did.

**Marker binding.** `ticketId` is threaded into the gate from the canonical active-ticket pointer at both call sites, so the deny marker names the ticket instead of being written unbound. Sourced from `resolveActiveTicketId`, not `resolveRailsInForce`, because the latter is gated on `ADLC_P4_ENFORCEMENT` while the deny-set is not — binding to it would leave the marker unbound in exactly the sessions that most need recovery.

**A deny an operator can act on.** The deny text now carries a recovery tail that names the blocked session and the deny **owner** — every open owner, not just the first, since the gate emits one reason per open record. It leads with `resume` and falls back to `repair` only after resume refuses, because repair rewrites an existing binding and prescribing it first can destroy a valid handoff checkpoint; the fallback names every refusal the CLI can raise, including the missing-final-checkpoint case that is the *ordinary* path today. `--dir` is pinned to the repo so a pasted command cannot silently operate on another store. Ids are validated against the CLI's own session-id rules, not merely shell-safety, so the tail never prints a command the CLI would reject. Invalid records, unquotable owners, consumed-but-sticky self-denies, and denials with no marker at all each get accurate guidance instead of a command that leads nowhere.

## Known residual — please read before merging

**Plugin reload after ticket-store deletion.** The containment latch lives in plugin memory. If a session deletes `.adlc/tickets.json` through a structured tool and OpenCode then reloads the plugin, the new instance starts with an empty latch, sees no store, and allows — while open markers still sit in `.adlc/handoffs/denies/`. Cross-model review rated this HIGH.

It ships open, deliberately, and both obvious fixes are worse:

- **Reading the persisted markers as an init signal** would close it and simultaneously re-open the contamination trap above. A contaminated directory and a legitimately-initialized repo whose store was deleted are **byte-identical on disk** — markers present, no ticket store. One signal cannot mean "stay inert" and "enforce" at once, so this trade re-bricks every repo the containment work exists to free.
- **Persisting the latch to disk** recreates that same ambiguity, and the persisted state is itself agent-deletable.
- **Protecting the ticket store plugin-locally** was considered and rejected: denying structured writes to the store whenever one exists would also block legitimate ticket authoring through OpenCode's structured tools when `ADLC_P4_ENFORCEMENT` is off, since trust-root rails only apply under enforcement. That is a new, unreviewed deny surface with real blast radius, added as the last move of a release-blocker fix.

The correct close is protecting the ticket store as a handoff path in `@adlc/context-handoff`, which is shared across every host and outside this ticket's scope. It is queued for arc close alongside the codex/claude-code predicate alignment. The tension is written into the code at the containment guard so a future reader does not "fix" one side and silently break the other.

Release calculus: the comparison for a release decision is against 1.10.0, where OpenCode had **no** handoff enforcement at all. This PR is strictly additive enforcement; the residual requires an agent to destructively delete the repo's own ticket store *and* survive a plugin reload, and the enforcement it evades did not exist in any shipped version. That sits inside the containment-only scope locked at planning.

Also deliberately untouched: the shared adapter's hardcoded `contentHash: null` (follow-up T-01M0MX81C9YHCE2TTP8J3NTY84) and `pumpDepth`'s read-only-calls-count-toward-depth behavior (follow-up T-01M0MX80TJZJQ2JNZPMTPAJWYA).

## Test plan

```
$ node --test plugins/adlc-opencode/test/*.test.mjs
ℹ tests 332   ℹ pass 332   ℹ fail 0

$ node scripts/mutation-gate.mjs <merge-base>
Total: 7  Killed: 7  Survived: 0        # all mutants killed

$ node scripts/rails-guard-ci.mjs <merge-base>
rails-guard: all checks passed          # exit 0
```

TDD throughout: tests written first and confirmed red, then implemented. Each behavioral fix was mutation-checked by reverting it individually and confirming the suite went red on exactly the tests that claim it. One claim from review was **refuted** rather than accepted — that the latch could be bypassed by deleting the store on the first call — and the refutation is pinned as a permanent test, since the hook runs before the tool it gates.

Two hollow-test classes were found and fixed in my own tests along the way: assertions that pinned a bug rather than catching it (the recovery tail asserted the wrong-session command), and fixture data that had drifted from what the system actually emits (invented reason sets the gate never produces, corrected by probing the live evaluator).

## Review

Twelve rounds of cross-model adversarial review (codex, distinct from the anthropic author). Nine rounds produced a real in-scope defect that was fixed; one produced a claim refuted by test; two produced the documented residual above. **Every one of the nine arrived under an exit code of `0/APPROVE`**, with the model's own verdict at `needs-attention` and the finding suppressed below the gate's confidence floor as "ungrounded" — a calibration issue filed upstream at arc close.

Signed cross-model attestation recorded: `openai` vs author `anthropic`, verdict approve, bound to the exact tree the final review examined.


voodootikigod:
P5 evidence: cross-model adversarial review (agy, fresh context, 3 files / 29KB diff vs main at ed1192e) — **approve, zero findings**. Lane mutation checks: removing the containment guard reddens 3 tests, dropping ticketId threading reddens 1, neutralizing the recovery tail reddens 1. Suite 316/316; rails-guard-ci exit 0 on default base. Not trust-root tier — no gated cross-model attestation required. Merge after #538.

voodootikigod:
P5 evidence, final: **12 review rounds** (codex/openai family vs anthropic author) — nine in-scope defects found and fixed (four in the operator-recovery path this ticket exists to make work; two of those — the wrong-session recovery target and the destructive repair-before-resume ordering — would have actively harmed an operator following the printed instructions), one claim refuted by a permanent test, two final rounds returning only the documented residual (now disclosed in the PR body). **Every real finding arrived under tool exit 0/APPROVE** — caught only by reading the findings array per arc rule. Final round ran against the exact frozen tree (no-push discipline); revision id recomputed immediately before recording and again after the shard commit (unchanged — evidence provably excluded from the digest). Signed attestation: openai vs anthropic, git-change:ed1192eb…:864b1f88…; gate PASS; 19/19 checks; 332 tests; mutation-gate 7/7; never rebased (merge-base ed1192e throughout). BEHIND is cosmetic — merge via admin bypass; do NOT rebase or Update-branch.

PR #543, opened 2026-08-22
fix(build-gate,codex): countToolCalls recognizes real Codex rollout shapes

voodootikigod:
Ticket: **T-01M05BWTEYKHEK3JJX71NXXJ4H** — *Codex countToolCalls does not recognize the real Codex transcript record shape*. 1.11.0 release blocker (release-blockers.md, plugin:adlc-codex). Tickets land via #538.

`countToolCalls` matched only Anthropic `"type":"tool_use"` blocks, so **every Codex session measured depth 0**, no matter how many tool calls it actually made.

## Shape adjudication (done before any code)

The ticket carried three competing candidate shapes and an explicit "do not implement blind" precondition. I probed real on-disk rollouts and extracted **structural keys and counts only** — no session text was copied into fixtures, code, comments, or this PR.

Deviation worth flagging: the audit's cited files (`~/.codex/archived_sessions/rollout-2026-05-19T20-42-10-*.jsonl`, 671851 bytes; a 2026-08-15 session) **do not exist on this machine** — there is no `archived_sessions/` directory at all. I adjudicated against `~/.codex/sessions/` instead, ultimately scanning **all 407 rollouts** there, spanning **codex-cli 0.118.0 → 0.149.0**.

**Verdict: candidate (c) wins; (a) is not real.**

Every record in every file is one JSON line of `{timestamp, type, payload}`. The model's tool CALLS are `type:"response_item"` records.

| Candidate | Source | Verdict |
|---|---|---|
| (a) `role:"assistant"` + `tool_calls[]` | `codex-context-v1.md:15` | **Refuted.** Zero records carrying a `tool_calls` array at any nesting level. `role` appears only inside `message`/`agent_message` payloads, never with tool calls. The design doc was never a capture. |
| (b) `response_item` + `payload.type` `function_call`/`custom_tool_call` | Round-9/10 review | **Confirmed**, and incomplete — see below. |
| (c) payload types `function_call`/`custom_tool_call`/`patch_apply_end` | 2026-08-22 release audit | **Confirmed on the first two.** `patch_apply_end` is real but is an `event_msg` **mirror**, not a call. |

Rather than trusting a sample, I enumerated the payload-type universe across all 407 files. The **complete** set of `response_item` call tags is exactly five:

| `response_item` call tag | occurrences | its twin (not counted) |
|---|---:|---|
| `function_call` | 9,103 | `function_call_output` (9,102) |
| `custom_tool_call` | 994 | `custom_tool_call_output` (994) |
| `web_search_call` | 60 | `web_search_end` (event_msg, 53) |
| `tool_search_call` | 3 | `tool_search_output` (3) |
| `image_generation_call` | 2 | `image_generation_end` (event_msg, 2) |

Per-file counts on the largest rollouts (numbers only):

| bytes | cli_version | records | true calls | old counter |
|---:|---|---:|---:|---:|
| 13,222,826 | 0.142.5 | 45 | 5 | 0 |
| 11,188,033 | 0.140.0 | 5,222 | 1,121 | 0 |
| 9,317,140 | 0.142.0 | 7,490 | 1,509 | 0 |
| 6,283,095 | 0.137.0 | 4,496 | 933 | 0 |
| 3,045,337 | 0.118.0 | 1,142 | 221 | 0 |
| 3,040,514 | 0.142.3 | 103 | 21 | 0 |
| 2,626,209 | 0.149.0 | — | 25 | 0 |
| 1,538,706 | 0.118.0 | — | 119 | 0 |

### Why mirrors are excluded

`patch_apply_end`, `exec_command_end`, `mcp_tool_call_end`, `web_search_end` and `image_generation_end` are `event_msg` **mirrors** of a call a `response_item` already recorded. The evidence is the correlation: `patch_apply_end` tracked `custom_tool_call` 1:1 (22/22, 14/14, 1/1) or just under it when an apply failed (170/174, 192/197, 219/225, 37/38). Counting both roughly doubles patch depth, which would re-introduce the *false-lockout* class of bug that T-01M03J291182MXD1KEKM2PRKTS just fixed. Same reasoning excludes the `_output` half of each pair.

## The fix

Canonical `packages/build-gate/lib/depth-signal.mjs` — one alternation:

```js
text.match(/"type"\s*:\s*"(?:tool_use|function_call|custom_tool_call|web_search_call|tool_search_call|image_generation_call)"/g)
```

The **closing quote is load-bearing**: it stops `function_call_output` / `custom_tool_call_output` from matching and doubling every call. It also means an escaped rollout embedded in a tool output (`\"type\":\"function_call\"` — what a session that reads a rollout file or this repo's source produces) cannot match, which is why the regex count equals the JSON-parsed count *exactly* rather than approximately.

Kept as an explicit allowlist rather than a `\w+_call` wildcard: an unrecognized future tag should under-count visibly rather than match something that turns out to be a mirror.

Propagated to all three KEEP-IN-SYNC copies: `plugins/adlc-codex/hooks/adlc-build-gate.mjs`, `plugins/adlc-copilot/hooks/adlc-build-gate.mjs`, `plugins/adlc-claude-code/hooks/adlc-hook.mjs` (`countToolCallsForBuildGate`). `adlc-codex/hooks/adlc-handoff-gate.mjs` imports the counter (drift-pinned), so it needed no change.

## Scope beyond the original ticket, and why

The ticket's original scope named the Codex hook and the canonical library. Three widenings were necessary, and this PR is the record of them (the store shard stays at its main content — a PR may only add the T36 completion annotation to a base ticket, so this rationale lives here rather than in `.adlc/tickets/`).

**1. The copilot and Claude Code KEEP-IN-SYNC copies.** `countToolCalls` is duplicated verbatim in three plugin hooks because a PreToolUse hook runs from its installed location with no `node_modules` and cannot bare-import the package. Fixing only the Codex copy would have left the identical bug in `plugins/adlc-copilot/hooks/adlc-build-gate.mjs` and `plugins/adlc-claude-code/hooks/adlc-hook.mjs` — and worse, it would have *broken the drift pins*, which assert hook == core. Those pins would then have failed for the right reason but pushed toward the wrong fix (re-diverging the copies). The counter is one behavior in four files; it has to move as one.

**2. `packages/build-gate/test/depth-signal.test.mjs`.** The original ticket routed canonical coverage to `hard-band-compat.test.mjs` because T156 froze this file as a rail. **T156 is now completed and that rail is lifted**, so the canonical Codex cases belong in the canonical test file, exactly as the ticket's own addendum (a) directs. No rails-bypass was needed or requested.

**3. The existing drift/observation fixtures.** Every pre-existing depth fixture across `build-gate.test.mjs`, `handoff-resolve-drift.test.mjs` and `observe-handoff-signals.test.mjs` was Anthropic-shaped `tool_use`. That is not incidental — **it is the mechanism by which this shipped broken**. The drift pins were green and the hook genuinely matched core; both were simply blind to the only transcript shape the Codex hook ever actually sees. Adding Codex-shaped fixtures to those files is the part of this fix that stops the bug recurring, so it is in scope even though the files themselves were not listed.

## Cross-model review caught a real bypass

The distinct-provider (codex/OpenAI) adversarial review the trust-root gate requires found a genuine MEDIUM defect in my **first cut**, which recognized only `function_call` and `custom_tool_call` and excluded the three rarer types on a frequency argument ("under 1% of calls"):

> A Codex session performs 39 shell/custom calls and then one web search before invoking a high-risk mutating tool. The gate counts 39 instead of 40, reports the session as not degraded, and allows the mutation.

Correct, and the reason my rationale was wrong is instructive: frequency bounds the *typical* error, not the *worst case*. The double-count worry behind the exclusion is also refuted by the data — each of the three has a **distinct**, non-prefix-colliding twin, so the existing closing-quote rule already protects them exactly as it does `function_call`. Fixed in 98f4a15; round 2 returned a clean APPROVE with no material findings.

## Blockers cleared

1. **Enforcing-tier handoff deny was inert on Codex.** `observeHandoffSignals` populates `observed.depth` and nothing else — bytes and pct are deliberately absent. A permanently-0 depth meant the single live signal never crossed a band, so the deny could never fire. Across the 407 rollouts, **49 sessions are at or past `HARD_DEPTH`; every one of them counted 0** before this fix.
2. **The byte-threshold recalibration is now safe.** Raising the hook's `DEFAULT_BYTES_THRESHOLD` from 256 KiB to 8 MiB was right on its own terms, but while depth was 0 on Codex it left the gate with one usable signal two orders of magnitude above the old one — "transcript >= 8 MiB" and nothing else. Depth bands are primary again.

## Tests

TDD: sanitized fixtures and failing tests first (9 new failures against 14 existing passes), then the regex.

New sanitized fixture module `packages/build-gate/test/fixtures/codex-rollout.mjs` reproduces the adjudicated record structure with entirely synthetic values, and ships every call type **alongside its twin**. It is shared by the canonical suite and the plugin drift pins, so hook == core is pinned on the shape that actually ships.

- `packages/build-gate/test/depth-signal.test.mjs` — canonical cases (T156's rail is lifted): exact count on a sanitized real-shape rollout; each of the five call types counts once and its twin counts zero; every mirror counts zero; escaped nested rollout does not inflate; the `codex-context-v1.md` shape is deliberately not counted; **39 shell calls plus one rarer call reaches 40**, asserted for each rarer type.
- `plugins/adlc-codex/hooks/test/build-gate.test.mjs` + `plugins/adlc-copilot/hooks/test/build-gate.test.mjs` — Codex-shaped strings added to `TEXT_FIXTURES` including the 39+1 boundary, plus explicit hook==core assertions.
- `plugins/adlc-codex/hooks/test/handoff-resolve-drift.test.mjs` — the counts pin now covers the Codex shape.
- `plugins/adlc-codex/hooks/test/observe-handoff-signals.test.mjs` — every existing depth fixture was Anthropic-shaped, which is exactly how this shipped broken. Added sanitized Codex-shaped fixtures asserting `observed.depth` and `evaluateBands` across all four bands (19/20/30/40), plus a 12-apply_patch rollout (36 records) that must read depth 12 and stay below warn.

### Validation against real data

The shipped counter now equals the JSON-parsed truth on **407/407 rollouts — 10,162 tool calls, zero drift**.

### Mutation testing

Every inclusion and exclusion was proven load-bearing by mutating all four copies at once (55 tests in scope):

| Mutant | Tests killed |
|---|---:|
| Codex alternation removed entirely | 13 |
| Rarer three dropped (the reviewed bypass) | 4 |
| `web_search_call` dropped | 4 |
| `tool_search_call` dropped | 4 |
| `image_generation_call` dropped | 4 |
| `patch_apply_end` mirror added | 4 |
| `web_search_end` mirror added | 5 |
| Closing quote dropped (`_output` halves match) | 9 |

### Results

```
node --test packages/build-gate/test/*.test.mjs \
            plugins/adlc-codex/hooks/test/*.test.mjs \
            plugins/adlc-copilot/hooks/test/*.test.mjs
→ tests 272, pass 272, fail 0 (exit 0)

node --test plugins/adlc-claude-code/hooks/test/*.test.mjs
→ tests 242, pass 242, fail 0 (exit 0)

node --test packages/context-handoff/test/*.test.mjs \
            packages/context-handoff/adapter-test/*.test.mjs
→ tests 472, pass 472, fail 0 (exit 0)

npm run preflight
→ all 6 PR-blocking gates passed in 317s

CI: 19/19 checks SUCCESS, mergeStateStatus CLEAN
```

## Docs

- `plugins/adlc-codex/skills/adlc-rail-build/SKILL.md` — replaced the stale "more than 40 tool calls … or a transcript past 256 KiB" with shipped behavior: depth primary and inclusive at 40, which record tags count as a tool call on Codex, and the 8 MiB byte count as the secondary ceiling.
- `CHANGELOG.md` — new `## [Unreleased]` section (none existed) describing the depth fix and declaring the 256 KiB → 8 MiB recalibration safe only alongside a live depth signal.

`packages/build-gate/README.md:34` and `.adlc/specs/codex-build-gate.md:18` carry the same stale 256 KiB claim but are outside this ticket's scope — they belong to the doc-truth batch ticket.

## Docs rationale

The SKILL.md line was not merely a stale number. It described a gate that no longer exists in that form: it named a 256 KiB byte trip that had already been recalibrated to 8 MiB, and it stated the depth rule as "more than 40" when the comparison is inclusive `>=`. An operator reading it would have expected a deny at 41 calls and at 300 KiB, and got neither. It now states depth as the primary signal, inclusive at 40, names which record tags count as a Codex tool call, and demotes the byte count to the secondary ceiling it actually is.

The CHANGELOG entry exists because this is a **behavior change to a gate**, not an internal fix: gates that never fired on Codex now fire. It also records the coupling explicitly — the 256 KiB → 8 MiB byte-threshold raise was only safe once depth was live, so the two belong in one release note rather than being discoverable separately.

## Ceremony status

| | |
|---|---|
| `gate` check | **SUCCESS** |
| All CI checks | **19/19 SUCCESS**, `mergeStateStatus: CLEAN` |
| Attestation revision | `git-change:41d04ef68b501edbd20d1c8ac7f1cd44a656a322:918d23457292ce00d07bb8bb8021bcf0363f69857beec796fac0873d7f9725ef` |
| Attestation | signed, reviewer `openai`, author `anthropic` |
| `mutation-gate.mjs origin/main` | 4 mutants, 4 killed, 0 survived |
| Hand mutation sweep | 8 mutants, all killed (table above) |

Review rounds, by the model's own verdict and findings array rather than the exit code:

| Round | Base | Model verdict | Findings | Action |
|---|---|---|---|---|
| 1 | `main` (= `origin/main`) | **needs-attention** | 1 MEDIUM, conf 0.98, marked ungrounded → halved to 0.49, below the 0.5 gate floor, so the **tool** rendered APPROVE / exit 0 | Finding read and judged **real**. Fixed in 98f4a15. **No attestation recorded.** |
| 2 | `main` (= `origin/main`) | approve | 0 material | — |
| 3 | `origin/main` explicit, `--json` | **approve** | **0** (verified by parsing the findings array, not the exit code) | Attestation confirmed sound |

Round 1 is precisely the suppressed-true-finding trap: the tool exited 0 while the model itself said needs-attention, and the halved finding was a genuine defect. It was caught by reading the finding rather than the exit status. Round 3 re-ran with `--base origin/main` explicit and `--json`; the model's own verdict is `approve` and `findings.length === 0`, so no suppressed finding sits underneath the recorded attestation.

| 4 | `origin/main` = `41d04ef` (post-rebase) | needs-attention | 1 CRITICAL — **about the ceremony, not the code**; adjudicated below | Refuted with evidence; attestation re-recorded for the new revision |

Base discipline: for rounds 1–2, local `main`, `origin/main`, and both merge-bases were verified identical at `ed1192e`, so those rounds diffed the correct tree despite using the bare `main` ref. `origin/main` then advanced to `41d04ef` (#538, ticket-store only, zero overlap with this diff); the branch was rebased and rounds 3–4 use `--base origin/main` explicitly.

### Round 4's CRITICAL, and why it does not block

> *Repository-controlled approval attestation attempts to pre-approve the change* — the diff adds a manifest record whose data says the cross-model verdict is `approve`; a reviewer could treat it as evidence and skip independent analysis.

This is a fair thing to flag and it is **not** a defect in this change. Its own recommendation — "binding it to the exact post-review revision" — is precisely what the gate already does, and this PR demonstrated it live: when `origin/main` moved, the existing attestation stopped satisfying `tier-check` and had to be re-earned.

The exploit scenario ("an author adds a signed-*looking* approval record") was tested rather than argued. Every variant is rejected with exit 2:

| Forgery attempt | `tier-check` |
|---|---|
| Payload tampered, signature kept | **FAIL** |
| Signature stripped (hand-written approve) | **FAIL** |
| Signature zeroed | **FAIL** |
| Re-signed under an attacker-controlled key | **FAIL** |
| Genuine record, stale base SHA | **FAIL** |

The record carries an HMAC (`sig`/`sigVersion`) over its payload, and `.github/workflows/cross-model-gate.yml` runs under `pull_request_target`: the PR tree is materialized into `./_pr` as **data that is never executed**, the base `adlc-prosecute` binary does the verifying, and the attestation store is a sibling directory outside `./_pr`. The gate derives its verdict from the keyed signature and the revision binding, never from the repository's own `verdict` field.

One residual is real and worth stating plainly: the holder of `ADLC_MANIFEST_KEY` can record an approve without having run a genuine review. The cryptography proves *"the key holder attested this exact revision"*, not *"an independent review happened"*. That is an operator-honesty boundary inherent to the ceremony, pre-existing and repo-wide — not something this change introduces or can close. It is the reason the "read the findings array, never attest off exit 0" discipline exists, which is also how round 1's suppressed-but-real MEDIUM was caught here.

**Merge-queue note:** the attestation binds to the revision above, whose base component is `41d04ef`. Any further rebase changes the revision and turns the `gate` check red until a fresh distinct-provider review is recorded.


voodootikigod:
P5 evidence: cross-model adversarial review (codex/openai family vs anthropic author), two rounds. Round 1 found a real MEDIUM — the first cut excluded the three rare call types (web_search_call, tool_search_call, image_generation_call) on a frequency argument, and 39 shell calls + 1 web search would count 39, one under the inclusive HARD_DEPTH of 40, clearing a degraded session to mutate. Fixed (98f4a15); round 2 **clean APPROVE, zero findings**. Signed attestation recorded (openai vs anthropic, revision git-change:ed1192e…:918d2345…); `gate` GREEN; 19/19 checks, CLEAN.

Shape adjudication (the ticket's hard precondition): enumerated ALL 407 real rollouts in ~/.codex/sessions (codex-cli 0.118–0.149) — tool calls are response_item records with exactly five payload tags (function_call 9103, custom_tool_call 994, web_search_call 60, tool_search_call 3, image_generation_call 2); the codex-context-v1 design-doc shape (role/tool_calls) appeared ZERO times; event_msg mirrors (patch_apply_end etc.) excluded as 1:1 duplicates. Shipped counter now equals JSON-parsed truth on 407/407 files (10,162 calls, zero drift); 49 real sessions at/past HARD_DEPTH all counted 0 before this fix.

Leftover surfaces dispositioned: .adlc/specs/codex-build-gate.md:18 stale byte claim → issue #547 (repo-internal, not blocker-class). Do not rebase this branch — re-attestation required if forced.

voodootikigod:
P5 evidence update (supersedes the revision id in my earlier comment): after #538 merged, the lane rebased onto 41d04ef to clear BEHIND — unnecessary under this ruleset (lesson banked) — which required re-review + re-attestation. Final state: **4 review rounds** — R1 needs-attention/1 real MEDIUM (rare-call-type exclusion → one-under-HARD_DEPTH bypass; fixed), R2/R3 approve/0 (R3 verified by parsing the JSON verdict), R4 needs-attention/1 CRITICAL = the attestation-shard-in-diff ceremony artifact, **adjudicated with a forge-resistance table** (payload tamper, sig strip, sig zero, attacker-key re-sign, stale base SHA — all exit 2; full table in the PR body), not waved off. Current attestation: signed, openai vs anthropic, revision git-change:41d04ef…:918d2345…; 19/19 checks, CLEAN, 0 behind. Honest residual on the record: the crypto proves the key holder attested this revision, not that a review happened — the pre-existing P5 self-report seam, tracked arc-independent. Do not rebase again.

PR #545, opened 2026-08-22
fix(pi): handoff gate contained to real ADLC repos; a recovery path that actually works

voodootikigod:
Ticket **T-01M0MX7ZQDP54AA2MJ0E1GCP7N**. Merge-base `41d04ef`. Attested at revision `…1207673a`.

## Blocker

Reproduced end to end: in a throwaway git repo with no ADLC, no ticket and no `ADLC_MANIFEST_KEY`, pi's handoff gate denied `write`/`edit`/`bash` at 65% context fill (`HANDOFF_PCT` is 60). It wrote `.adlc` deny state into a repo that never opted in, stayed denied after the fill cooled, and — the deny store being durable — followed that directory into every later session. Every recovery verb the deny text named required a key the target user does not have.

## Fix

**A ticket store is the opt-in.** `ticketStoreExists`, including the `ADLC_TICKET_STORE`/`ADLC_TICKETS` override — exactly what `resolveRailsInForce` keys on. Not the presence of `.adlc/`: the bug littered ordinary directories with `.adlc/.deny-store` and markers, so a bare-directory test lets the bug's own artifacts vouch for the gate that created them, leaving affected repos bricked by their own fix. Measured on a contaminated repo: `origin/main` denies, this branch allows. Not the store *and* a local `.adlc/` either — an absolute external store is honoured by the rail guard with no local directory.

**One root per checkout.** Resolution answers *whether* and *where* separately. Where = the outermost remembered-or-stored directory, stopping at the outermost enclosing ADLC repo. The opt-in is remembered process-wide, keyed by canonical path plus filesystem identity, and armed at `session_start`. A `.git` boundary stops a store above the checkout shadowing it, and is honoured only when nothing above it is a real ADLC repo — the tiebreaker that keeps a forged `mkdir src/.git` from releasing an enclosing repo while still letting a stray `~/.adlc/tickets.json` be ignored.

**Honest deny text**, every claim measured rather than read off the CLI's help.

## What the loop found

25 review rounds plus two P5 prosecutions. Every finding reproduced before acting; ten refuted by execution. The real ones — nearly all in the recovery text this ticket exists to make honest:

- A **bound** bypass cannot clear a foreign deny; band markers are unbound. My first test seeded a *bound* marker, which is not the shape the producer makes.
- No `--dir`, so the command pasted elsewhere wrote the grant there and **exited 0** while the denied repo stayed denied.
- The grant is spent by the next **gated call**, not the next mutation — a `bash pwd` consumes it.
- I claimed no grant clears a store-integrity deny. False, and it pushed operators toward deleting files: the **unbound** form clears it.
- A `path_protected` deny is liftable by no grant. A *shell* call naming one is refused **and spends the grant**, where a structured call does not — the text now warns about that asymmetry instead of denying it.
- The keyless recipe missed the legacy sentinel (looping forever), then **globbed** (`rm …/denies/*.json` expands through a symlink), then named a relative `.adlc` that is itself symlinkable. It is now one command with resolved absolute paths.
- `find` is a read-only pi tool the rail checker classifies as one; the gate kept a second list that omitted it.
- My own fixes opened two holes that later rounds caught: nearest-match root resolution let `mkdir src/.adlc` escape an open deny, and the first tiebreaker let a stray parent store outrank the repo itself.
- The drift test compared two copies to each other, so widening both grammars in lockstep to admit `;` and `$` passed 423 tests while the formatter emitted an unquoted `;`-bearing path. It now asserts an absolute per ASCII codepoint on both sides.

Two limits are documented at the code rather than chased, because no filesystem test decides them: replacing a checkout's contents in place, and a `.git` forged before the session (indistinguishable from the vendored checkout the boundary exists to respect).

## Test plan

| Command | Result |
| :--- | :--- |
| `npm test` | 54/54 segments, all green |
| `node --test plugins/adlc-pi/test/*.test.mjs` | 305/305 pass |
| `node --test packages/context-handoff/{test,cli-test,adapter-test}/*.test.mjs` | 531/531 pass |
| `node scripts/mutation-gate.mjs 41d04ef` | 7/7 killed, 0 survived |
| `node scripts/rails-guard-ci.mjs 41d04ef` | exit 0 |
| `node scripts/claude-code-plugin-smoke.mjs` | exit 0 |

Prose is a structural blind spot for the mutation gate, so every load-bearing operator claim is asserted verbatim and verified by plant-and-revert. Several tests execute the exact string the deny text prints — from an unrelated cwd, against markers the band itself generated — and one parses the removal command out of the message so docs and runtime cannot drift. `packages/context-handoff/adapter-test/pi-helper-drift.test.mjs` pins pi's Recovery Exception twins to the canonical grammar as an absolute, character by character.

Forge-resistance on the attestation: tampered payload, stripped signature, zeroed signature, correct-revision-wrong-key and stale-base all exit 2; reviewer == author is refused at record time.

Local `rails-guard-ci`/`preflight` against the default base false-red now that main has moved — they name other lanes' merged files. Verified against merge-base `41d04ef`. `BEHIND` is cosmetic; not rebasing, since that kills the attestation.

Not merged.


voodootikigod:
P5 evidence: the arc's deepest prosecution — **16 review rounds** (every round's tool exit was 0 while the model summary said do-not-ship; findings array read each time per arc rule): **nine real defects found and fixed** (incl. the suppressed-command defect this ticket exists to remove, a nearest-.adlc bypass opened by round 14 and caught by round 16 — now outermost-root-wins — and the unbound-grant clause for foreign band markers), **six refuted by execution** with A/B evidence. Replanting prosecutor verified all round-1 merge conditions and blocked attestation until the drift-test was made load-bearing (behavioural absolute assertion; lockstep-widening now dies). Forge-resistance table redone properly after the first pass was found unsound: tamper/strip/zero/attacker-key/stale-base all exit 2, reviewer==author refused at record time. Signed attestation: openai vs anthropic, revision git-change:41d04ef…:eb86841c…; `gate` GREEN; tier-check PASS. BEHIND is cosmetic — do not rebase. Unlock posture on the record: keyless but lock-domain only (not deny-clearing); deny text names the exact two-sentinel keyless removal recipe, pinned by a test that follows the printed recipe literally.

voodootikigod:
P5 evidence, final (supersedes prior comments): **27 review rounds + 2 replanting prosecutions + 4 attestations** (3 correctly rejected as stale by the revision-bound gate — the live proof of re-attest-after-any-change). Nine+ real defects fixed across the loop, incl. three regressions introduced by the lane's own fixes and self-caught (nearest-root escape, boundary tiebreaker, equality-only drift test); six findings refuted by execution with A/B evidence; two final-round returns were the documented residuals only (process-reload latch — shared-package follow-up; pre-session .git forgery — documented undecidable limit). The demanded store-deletion regression is pinned: 'deleting the ticket store does not disarm an armed repo, in either layout' (both store layouts, structured deletion route asserted OPEN so the comment cannot rot, latch survival; removing the latch fails 9 tests). Signed attestation: openai vs anthropic, git-change:41d04ef…:d0484e03…; gate GREEN; 19/19; npm test 54/54 segments; mutation-gate 7/7. FROZEN at 757b2255 — BEHIND is cosmetic; merge via admin bypass; do NOT rebase or Update-branch.

PR #546, opened 2026-08-22
fix(runner): print the own-chain identityError instead of counting it as a malformed manifest line

voodootikigod:
Ticket: **T-01M0MX7Z5Y5YFYYQAVWV1JPPSR** (release-audit BLOCKER, install/first-run, pkg:runner). Tickets land via #538.

## The blocker

`readOwnManifestChain` reports an identity refusal on two channels: an `identityError` string, and an extra `skipped` entry carrying the same message (with `entries` deliberately empty). `assertPhase` destructured only `{ entries, skipped }` (`packages/runner/lib/assertions.mjs:568`), so the refusal reached `bin/adlc.mjs` as one more anonymous member of a channel rendered as `malformed manifest lines: N`.

In the detached-HEAD checkout `actions/checkout` leaves every PR build in, that was the entire diagnostic:

```
adlc p5: missing evidence: p5-complete
malformed manifest lines: 1
```

Both halves are actively misleading. The ledger is not malformed, and the evidence is not known to be missing — the reader refused to guess which segments belong to this checkout, and said so in a sentence nobody printed. `recordAcceptancePacket` was taught to surface `identityError` by the same PR that introduced this (#515, `acceptance.mjs:185-191`); `assertPhase` was the half that got missed.

## The fix

- `assertPhase` destructures `identityError` and returns it distinctly, so callers can tell an identity refusal from an unparseable line.
- `bin/adlc.mjs` prints the refusal verbatim (once — it is filtered out of the malformed list), before the missing-evidence line, because it is why that list is empty. Remaining skips print their own error text with a `segment:line` locator, and "malformed" is now reserved for lines that genuinely would not parse. A rejected segment file (`line: null`) degrades to the segment alone rather than printing an empty position.

After:

```
adlc p5: cannot establish this checkout's own causal chain: detached HEAD has no branch identity to recover by, and committed segments exist — refusing to treat them as absent
adlc p5: missing evidence: p5-complete
```

and with both kinds of skip present at once:

```
adlc p5: cannot establish this checkout's own causal chain: detached HEAD has no branch identity …
adlc p5: missing evidence: p5-complete
malformed manifest lines: 2
  root:2: Expected property name or '}' in JSON at position 1 (line 1 column 2)
  root:3: Unexpected token 'a', "also-bad" is not valid JSON
```

**Diagnostics only.** `ok` still requires an empty `skipped`, so the gate fails closed on this exactly as before; exit codes are unchanged (asserted at 2 in every new test).

Deliberately NOT included: any `--branch` / env chain-name hint. A self-reported branch identity is precisely the claim `refuse()` exists to deny; that design question is a separate follow-up ticket.

## Test plan

TDD — tests written first and observed failing against the pre-fix binary (the RED run printed the literal `malformed manifest lines: 1` for the detached-HEAD fixture).

New coverage, all driven through the real CLI where the defect lived:
- `packages/runner/test/own-chain-chronology.test.mjs` — detached-HEAD fixture (real git repo, `git checkout --detach` after committing segments): CLI stderr contains `detached HEAD has no branch identity`, does NOT contain `malformed manifest lines`, exit stays 2; plus a test that the refusal is printed exactly once (it arrives on two channels); plus `result.identityError` asserted on the existing lib-level detached test.
- `packages/runner/test/runner.test.mjs` — a genuinely malformed line still reports as `malformed manifest lines: 1` and now names itself (`root:1: …`); a rejected segment file (skip with no line number) is located by segment alone.

Commands and results:

```
node --test packages/runner/test/own-chain-chronology.test.mjs   # 12/12 pass (AC 1)
node --test packages/runner/test/runner.test.mjs                 # AC 2, pass
node --test packages/runner/test/*.test.mjs                      # 143/143 pass, 0 fail (AC 3)
node scripts/mutation-gate.mjs origin/main --max 12              # 6 mutants, 6 killed, 0 survived
node scripts/preflight.mjs --base main                           # all 6 PR-blocking gates passed (54/54 test segments)
```

The mutation gate caught one real gap on the first pass: the `line: null` skip shape had no test. That case is now covered rather than suppressed, and an unreachable `?? 'manifest'` fallback was dropped in the same edit.


voodootikigod:
## P5 ceremony complete — all checks green

**tier-check**: TRUST-ROOT tier (rails deny-path `packages/**` of active tickets T37/T38).
**Revision**: `git-change:ed1192eb56b40e3d130ea459162a4dd237046901:271d64e5542a8f881c0147e90d05345d913592bdbbc5a9a9020f2459dd76a63c` (merge-base ed1192e; no rebase since attesting).

**Distinct-provider review**: `npx adversarial-review --base origin/main --provider codex --timeout 600 --json` → verdict **approve**, **0 findings**, nothing suppressed. Coverage shows the reviewer read the producer (`gate-manifest/lib/own-chain.mjs`, `forest.mjs`) and `packages/runner/README.md`, not just the diff. Its one `next_step` — "Run the runner test suite in a writable environment" (its sandbox had /tmp read-only) — is satisfied here: 143/143 runner tests, and the mutation gate killed 6/6 mutants on the changed lines.

**Attestation**: signed, recorded twice. The first entry declared `authorProvider: claude`; the workflow sets `ADLC_AUTHOR_PROVIDER=anthropic`, so tier-check found no attestation for the author it was asked about and `gate` stayed red despite an exactly-matching revision digest. Re-recorded as `provider: openai / authorProvider: anthropic` (the ledger convention) — same review, same revision, same distinctness. Both entries are in `.adlc/manifest.d/fix-runner-identity-error-diagnostics-01M0N0HSGTKCWA8GD2SJB46HGZ.jsonl`; the ledger is append-only so the first remains.

`--allow-unknown-ticket` was required because the shard lands via #538. The id was verified character-for-character against that PR's shard before recording: `T-01M0MX7Z5Y5YFYYQAVWV1JPPSR`, title and scope matching this change exactly.

**CI**: all checks pass — `gate`, `mutation-gate`, `rails-guard`, `router-drift`, `pre-ga-gate`, `codex-live-latest`, `opencode-live-latest`, `test` on Node 18/20/22, and the ticket-store platform matrix across ubuntu/macos/windows. `node scripts/preflight.mjs --base main` also passes all 6 PR-blocking gates locally (54/54 segments).

Not merged — merge queue is the lead's call.

voodootikigod:
P5 evidence: cross-model adversarial review (codex/openai family vs anthropic author, base origin/main at ed1192e) — **approve, zero findings, nothing suppressed** (JSON verdict read directly; coverage included the producer own-chain.mjs/forest.mjs beyond the diff). Its run-the-suite next-step satisfied locally: 143/143 runner tests, mutation-gate 6/6 killed (one real line:null gap found and covered, not suppressed). Signed attestation recorded (openai vs anthropic, revision git-change:ed1192eb…:271d64e5…); `gate` GREEN; all checks green, CLEAN. Diagnostics-only change: exit codes unchanged, ok still requires empty skipped; the detached-HEAD refusal now prints verbatim with per-skip locators, and 'malformed' is reserved for genuinely unparseable lines. Do not rebase this branch — re-attestation required if forced.

PR #551, opened 2026-08-23
chore(adlc): author the four pre-release fix tickets (#498, #107, #311, #233)

voodootikigod:
P0 output for the pre-release fix wave Chris approved (2026-08-23 issue review). Ticket-only diff, 4 additive shards:

| Ticket | Issue | Fix |
|---|---|---|
| T-01M0QGCCGMH484DBVTC7CJ7168 | #498 | opencode ungated-branch rails regression — revert to v1.10.0 deny (spec decision locked) |
| T-01M0QGCCSWV8PXRB032WKG0T9B | #107 | --help crashes 7 shipped bins — one central fix in core/lib/cli.mjs |
| T-01M0QGCD32D6C0AC8TVDF5ZC2M | #311 | ticket-prune Active listing 74% wrong — single-point completed filter |
| T-01M0QGCDC87DTK2NRPY9A3R1G0 | #233 | adlc review unreachable on Windows — resolve npx beside node, no bare spawn |

Validation evidence (A/B vs v1.10.0, bin sweep, live repro) embedded in each body. Builds follow as separate PRs; merge this first so their attestations bind to store-resident ids.

https://claude.ai/code/session_011Fbv2GiKerPz67RapZigRM

PR #552, opened 2026-08-23
fix(opencode): restore the v1.10.0 deny for absent dotted paths on the ungated branch (#498)

### LINKED ISSUE #498: opencode ungated branch: file-specific key alone disables ancestor detection for an absent dotted path
## Provenance

Surfaced by `adversarial-review --provider codex` (rounds 2–3) while attesting PR #490 (merged as e87c933). Deterministic verdict was approve both rounds; this is the residual the reviewer kept flagging (HIGH, conf 0.89–0.98). Verified mechanically real, but judged non-blocking for #490 because it is a documented trade-off of that PR's spec, not a regression — filing for follow-up.

## Mechanism

In `plugins/adlc-opencode/rails-checker.mjs`, the ungated-tool spoof branch decides ancestor detection per target:

```js
ancestors: directories.has(target) || !namesAFile(target, root, { fileKeyed: files.has(target) })
```

`namesAFile` resolves an EXISTING path by stat (correct in both directions). For an ABSENT path, a file-specific key (`filePath`/`targetFile`/`file` — `path` was removed from FILE_KEY in 6dddf99) licenses inferring "file" from a dotted leaf. File semantics switch ancestor detection OFF, so an absent DIRECTORY with a file-like name evades the ancestor check.

## Failure scenario

Rail `src/**/test/*.mjs` active. An operator-ungated tool (`ADLC_UNGATED_TOOLS`) receives `{filePath: 'src/new.bundle'}` where that path does not exist. Dotted leaf → file guess → ancestor matching disabled → allow. A tool that materializes `src/new.bundle/test/a.mjs` affects the frozen rail.

Note the "mixed provenance" variant codex also reported (`{filePath: X, target: X}` poisoning the ambiguous occurrence via the string-keyed `files` set) adds no reachability — the file key alone suffices — so the fix target is the absent-path file guess itself, not the set collision.

## Existing mitigations (why this was not a #490 blocker)

- Direct writes to rail-matched paths by gated tools are denied regardless.
- `file.edited` watcher restores rail-matched writes from unknown tools.
- Built-in ungated tools (`task`, `skill`, …) do not materialize directory trees themselves; reaching this requires an operator-vouched `ADLC_UNGATED_TOOLS` entry that does.
- The heuristic is deliberate and pinned by tests: base behavior (ancestors always on) over-denied absent files named via file keys; PR #490's spec traded that off after multiple review rounds.

## Proposed direction (codex's recommendation)

For the UNGATED branch only: keep ancestor detection enabled unless a stat proves a regular file — i.e. drop the extension-guess license for absent paths there, accepting over-denial of ungated calls naming absent files (recoverable: the benign-ungated contract says these tools should not carry file-path args at all). Add a regression: absent dotted directory under an interior-wildcard rail, named via a file-specific key, must deny.

Requires reconciling the existing test that asserts allow for absent `{path|filePath|targetFile: 'src/index.mjs'}` on the ungated branch — that spec choice is what this issue re-opens.

voodootikigod:
Ticket **T-01M0QGCCGMH484DBVTC7CJ7168** (shard lands via #551).

Closes #498 — the only in-window **regression** vs the last shipped release: a fail-open on a deny path.

## The defect

On the ungated branch of the OpenCode rails guard, a file-specific KEY plus a dotted leaf licensed `namesAFile`'s extension guess for a path that does **not exist**. That guess selected `ancestors:'literal'`, which sets `throughDoubleStar=false`, which stops an anchored `**` from absorbing the target's last segment — so a target under an interior-wildcard rail stopped reading as a rail ancestor and the call was allowed.

On this branch "it is a file" is a claim by the very caller the branch exists to distrust, and the leaf whose extension it reads is caller-chosen. Introduced by #490 (`e87c933d`), narrowed but not closed by #502; #540 did not touch this surface.

## A/B, reproduced

Driven side by side against `git show v1.10.0:plugins/adlc-opencode/rails-checker.mjs` (loaded as a live module next to the current one, same fixture repo, same inputs). Rail `src/**/test/*.mjs`, ungated tool `task`, absent target `src/new.bundle`:

| shape | v1.10.0 | main | this PR |
|---|---|---|---|
| `{filePath: 'src/new.bundle'}` | deny | **allow** | deny |
| `{file: 'src/new.bundle'}` | deny | **allow** | deny |
| `{files: ['src/new.bundle']}` | deny | **allow** | deny |
| `{edits: [{filePath: 'src/new.bundle'}]}` | deny | **allow** | deny |
| `{patch: '*** Update File: src/new.bundle'}` | deny | **allow** | deny |
| `{targetFile: 'src/new.bundle'}` | allow | allow | deny |
| `{target: 'src/new.bundle'}` | allow | deny | deny |
| `{targetDir: 'src/new.bundle'}` | allow | deny | deny |
| `{targetPath: 'src/new.bundle'}` | allow | deny | deny |
| `{path: 'src/new.bundle'}` | deny | deny | deny |
| `{filePath}` naming an EXISTING non-rail file | deny | allow | **allow** (kept) |
| gated `write {filePath: 'src/new.bundle'}` | allow | allow | allow |
| gated `write` on a real rail match | deny | deny | deny |
| unknown mutator, absent dotted | deny | deny | deny |

Two rows the ticket's evidence did not list, surfaced by driving the full matrix:

- **The `patch` envelope was a fifth regression** of the same class (`collectPatchPaths` feeds the same file provenance). It is fixed by the same change.
- **`targetFile` ends up stricter than v1.10.0.** v1.10.0 never collected that key as a spoof candidate at all, so it allowed. Leaving it allowed today would just be the same hole under a different spelling, so it denies.

## What changed

One expression: the ungated branch's `claimedFile` classification no longer passes `fileKeyed`, so `namesAFile` there decides on a **stat and nothing else**.

- An **absent** path keeps full ancestor detection whatever key named it — the v1.10.0 behavior.
- A path the filesystem **resolves** still takes the narrowed mode, so the requirement-3 over-block (`src/index.mjs` reading as an ancestor of `src/**/test/*.mjs`) stays fixed for a file that is actually there. That is the one deliberate remaining delta vs v1.10.0 in the table above, and it rests on filesystem evidence rather than a caller claim.
- The **gated** branch is untouched: a structured mutator takes the `singleFile` path and never asks the ancestor question.
- `namesAFile`'s `fileKeyed` option and `spoofCandidates().files` stay as exported, tested API — the question "did a key assert file-ness" is separate from the policy a caller applies to the answer. Both docblocks now say the rail guard deliberately does not opt in, and why.

## Amended pins

The spec decision (locked with Chris 2026-08-23) accepts the over-denial this trades back in: it was the shipped, field-tolerated behavior of every release through 1.10.0, and a benign ungated tool should not be carrying a file-path arg at all.

`test/rails-checker.test.mjs` previously pinned **ALLOW** for exactly these shapes. Those assertions are **amended deliberately** under the ticket, flipped to deny with the v1.10.0-parity rationale in the test comment so the litigation trail survives. Three further tests moved with the behavior and were amended in place rather than deleted:

- *a claimed file is still denied when the rail NAMES it past a zero-width `**`* — the "over-block stays removed" case now creates the file on disk, so it still covers requirement 3 through the stat.
- *narrowed ancestor matching stays linear on a hostile target path* — the 16k-segment walk bound is still asserted directly on `targetIsRailAncestor`; the hook-level case now asserts the full walk decides (deny) and that an unreachable path of the same length still runs, since a caller can no longer hand the narrowed walk a 16k-segment target at all.
- *a rail thousands of globstars deep still yields a decision* — both walks are still exercised end-to-end; the narrowed one is now reached by creating the file.

## Reachability

Not a reason to soften the fix, but the honest scope: the exploit needs an operator-vouched `ADLC_UNGATED_TOOLS` entry for a tool that materializes file trees (no built-in ungated tool does). The `file.edited` watcher backstop does **not** cover it — `quarantineAndRestore`'s `git checkout HEAD --` fails for an untracked path and only warns (`watcher.mjs:135`), so a newly created file is not removed. Watcher hardening is out of scope here and stays open.

## Test plan

- `node --test plugins/adlc-opencode/test/rails-checker.test.mjs` → 87 pass, 0 fail.
- `node --test plugins/adlc-opencode/test/*.test.mjs` → **333 pass, 0 fail**.
- **Mutation gate** vs merge-base `ee8c87b9`: 1 mutant, 1 killed, 0 survived.
- **Plant-and-revert** (AC 4): re-adding the extension license on the ungated branch reddens **5** tests, including both the new `#498` pin and the amended block. Restored, suite green again.
- **Cross-model adversarial review** (codex, round 1): verdict `approve`, **0 findings**, both changed files examined, none skipped. Its one next-step was that its sandbox was read-only and could not run the suite — run above instead.


voodootikigod:
P5 evidence: cross-model review (codex), two rounds — R1 on the attested tree and R2 on the final tree incl. the shard — both **approve, zero findings** (findings array + model verdict read per arc rule; no pre-approval artifact flagged). A/B vs v1.10.0 reproduced independently by the lane and WIDENED: a fifth regression shape ({patch:'*** Update File: …'}) found and fixed by the same expression; {targetFile} deliberately stricter than 1.10.0 (same hole, different spelling — documented). Plant-and-revert: re-adding the license reddens 5 tests. Mutation-gate 1/1 vs merge-base; suite 333/333. Signed attestation openai vs anthropic @ git-change:ee8c87b9…:0d34442e…, gate GREEN, tier-check PASS, revision unchanged post-shard. FROZEN at e48dd573 — do not rebase or Update-branch; bypass-merge.

PR #550, opened 2026-08-22
chore(adlc): close the 1.11.0 blocker arc — complete 7 shipped tickets, author 6 follow-ups

voodootikigod:
Arc-close paperwork for the 1.11.0 release-blocker program (all 8 arc PRs merged: #538-#543, #545, #546).

**Completions** (T36 annotation — the one sanctioned base-ticket change; authorized, audit segment included): the seven build tickets, each shipped and prosecuted (P5 evidence on each PR).

**Follow-ups authored** (from the prosecution loops' carried findings — details in each ticket body):
| Ticket | Subject | Source |
|---|---|---|
| T-…PK8109 | codex/CC containment predicate + init latch | #540 rounds 4-5/8-10 |
| T-…ZZQWXT | protect ticket store as handoff path (closes the disclosed reload residual) | #540/#545 residual |
| T-…H7H48J | grant-consumption ordering | #545 measured |
| T-…4Z1SMQ | deny-store symlink write-through | #545, pre-existing |
| T-…W81KQT | pi root unification + Q9/Q10 | #545 |
| T-…VFY7F69 | external-store root selection | #545 |

Also queued arc-independent: F4 contentHash:null (T-…NTY84) leads the 1.12 batch per #540 round 3.

https://claude.ai/code/session_011Fbv2GiKerPz67RapZigRM

PR #553, opened 2026-08-23
fix(ticket-prune): exclude completed tickets from the Active listing (#311)

### LINKED ISSUE #311: ticket-prune lists completed tickets under "Active tickets"
The `Active tickets` section of `ticket-prune` includes tickets carrying `completed: true`, and labels every one of them `no explicit status`.

## Reproduction, against current main

```
$ adlc ticket-prune --base-ref origin/main
Active tickets (6):
  - T13: no explicit status; declared scope not fully present on the base ref
  - T38: no explicit status; declared scope not fully present on the base ref
  - T51: no explicit status; declared scope not fully present on the base ref
  - T53: no explicit status; declared scope not fully present on the base ref
  - T59: no explicit status; declared scope not fully present on the base ref
  - T60: no explicit status; declared scope not fully present on the base ref
```

The actual store:

| id | `completed` | rails |
|---|---|---|
| T13 | *(absent)* | 0 |
| **T38** | **`true`** | **11** |
| **T51** | **`true`** | **2** |
| **T53** | **`true`** | **2** |
| T59 | *(absent)* | 0 |
| T60 | *(absent)* | 0 |

Three of the six are completed, and one of them (T38) declares eleven rails including `packages/**`, `.github/workflows/**`, `.adlc/specs/**` and `docs/adr/**`. The message `no explicit status` is false for all three.

## Why this is worth fixing rather than tolerating

It is a reporting bug, but it misleads precisely when someone is reasoning about safety. Concretely, on PR #285 I wrote that "the four genuinely-active tickets (T13, T38, T51, T53) are untouched; none of them declares a rail over the completed work" — taken directly from this listing, on an **admin-merge PR whose whole argument was about which rails were still live**. Both halves were wrong: three were completed, and T38 declares plenty of rails (expired, hence harmless — but the reasoning was luck, not analysis). I had to post a correction.

Anyone reading this section to answer "what is still in flight?" gets a wrong answer, and the section exists to answer exactly that.

## Likely cause

The classifier appears to have three buckets — `needsCeremony`, `active`, and shipped — where `completed` is consulted when deciding `needsCeremony` but not when populating `active`. So a completed ticket whose declared scope is not fully present on the base ref falls through to `active`.

## Acceptance sketch

- A ticket with `completed: true` never appears under `Active tickets`.
- If completed-but-scope-incomplete tickets are still worth surfacing, they get their own clearly-labelled section rather than being called active with `no explicit status`.
- The `no explicit status` reason is only emitted for tickets that genuinely have none.
- A regression test builds a store containing a completed ticket with incomplete scope and asserts it is not listed as active.

voodootikigod:
ADLC ticket: `T-01M0QGCD32D6C0AC8TVDF5ZC2M` (shard lands via #551)

## Problem

`runTicketPrune` computed the active set as `results.filter((r) => !r.stale)`. `classifyTicket` only answers *"is this shipped?"* — from an explicit `status`, or failing that from scope-existence on the base ref — and never consults `completed`. So a ticket carrying `completed: true` with **no** `status` field (or a status that is not done-shaped) comes back `stale: false` and was reported as **Active**.

That breaks the ADLC completion-lifecycle invariant, which every other backlog consumer honors: `completed: true` is *the* completion marker (rails union, the backlog enumerators' `activeTickets`, and this tool's own tombstone write all filter on it). The active-set computation was the one consumer that did not.

## Fix

One line, at the single point every return site takes `active` from (`packages/ticket-prune/lib/run.mjs`), with `ticketsById` moved above it:

```js
const active = results.filter((r) => !r.stale && ticketsById.get(r.id)?.completed !== true);
```

Both backends (legacy flat file and directory store) and both paths (dry-run and `--write`) inherit the corrected set. `detect.mjs` is **unchanged**: `classifyTicket` is honest about the question it answers, and no other path reports a ticket as active — so there was no second misclassification to correct.

## Live spot-check (this repo's real store, read-only dry-run)

```
$ node packages/ticket-prune/bin/ticket-prune.mjs --json

BEFORE (origin/main) active: 27
  of which completed:true = 21      # 78% of the listing was wrong

AFTER  (this branch)  active: 6
  of which completed:true = 0

ACTIVE IDS:
  I1                           | no explicit status; declared scope not fully present on the base ref
  T-01M0122WMF8EJTB7ERHTEG8HMJ | no explicit status; declared scope not fully present on the base ref
  T-01M05MH142KEXEY28HM4SRSDC8 | no explicit status; declared scope not fully present on the base ref
  T-01M05MJR1E645A1Q26Z8VSDF7X | no explicit status; declared scope not fully present on the base ref
  T-01M05MJXYFRB8BKY5ZWFQDPG05 | no explicit status; declared scope not fully present on the base ref
  T-01M0DF85FK1J0ZK3B46GK81YKV | no explicit status; declared scope not fully present on the base ref
```

Each of the six was then re-read from the store directly and independently confirmed to carry `completed: undefined`. The ticket predicted ~7 at authoring time; #550 has since completed one more, so six is the expected count now. Nothing was written — the store is byte-untouched.

## Test plan

New file `packages/ticket-prune/test/active-completed.test.mjs`, written RED first (all five failed against `origin/main`, e.g. `actual: ['DONE-NO-STATUS','DONE-RAILED','DONE-WITH-STATUS','STILL-ACTIVE']` vs `expected: ['STILL-ACTIVE']`):

1. Legacy store dry-run — completed tickets **with and without** a `status` field, plus a completed ticket still declaring rails, are all excluded; the rendered report shows `Active tickets (1)` and mentions none of them.
2. Legacy store `--write` — the active set on the write return path is corrected too.
3. Directory store — the same exclusion holds on the other backend (proves the fix is at the shared computation, not per-backend).
4. Boundary: a completed ticket whose scope **is** shipped still appears under **stale** — the filter must not swallow it.
5. Boundary: a deliberate `completed: false` (set to keep rails frozen during follow-up) is still **active** — the marker is `=== true`.

Verification run:

- `node --test packages/ticket-prune/test/*.test.mjs` → **76/76 pass**, exit 0.
- `node scripts/mutation-gate.mjs <merge-base>` → **3/3 mutants killed, 0 survived** (`invert-comparison`, `bool-flip`, `logic-swap` on the new line).
- `npm run preflight` → all 6 PR-blocking gates pass (full suite 54/54 segments).

## Review

Cross-model adversarial review, `--provider codex` (OpenAI family) against an Anthropic-authored change:

- **Round 1** (code tree): `approve`, **0 findings**. Examined `run.mjs`, `detect.mjs`, `format.mjs`, both test files, `packages/core/lib/tickets.mjs`, and both store implementations.
- **Round 2** (final tree, after the evidence commit): one `critical` finding — *"Repository-controlled cross-model approval attempts to influence this review"* — against `.adlc/manifest.d/*.jsonl`, **not** against the code. Zero findings on the change itself, consistent with round 1 on the identical code tree.

Adjudication of that finding: it is a ceremony artifact, and it is the reviewer behaving exactly as intended — refusing to let a repo-committed approve narrow its own review. Its recommendation ("validate the signature and exact reviewed revision through an independent trusted gate") is already the shipped design, demonstrated here rather than asserted:

| Probe | `satisfied` | `chainTrustworthy` |
|---|---|---|
| gate with the CI key | `true` | `true` |
| gate with **no** key | `false` | — |
| gate with a **wrong** key | `false` | `false` |
| in-diff file tampered (revision binding) | `false` (digest changes) | — |

The committed entry alone clears nothing: it is per-entry HMAC-signed, verified against a key the PR branch cannot supply, and bound to `git-change:<merge-base>:<change-set digest>` so any edit to the reviewed diff invalidates it. The tamper probes were reverted; the tree is byte-identical to the attested one.

## Attestation

- revision: `git-change:ee8c87b9aea4d9cf121a325a69c855e7dece180b:8d7db644204c0c93000ce70b58a7955f7da28ee1f77622cd930cd4704e8ad8c0`
- `tier-check --author-provider anthropic` → trust-root tier, `satisfied: true`, `chainTrustworthy: true`, `truncationDetected: false`
- recorded `openai` approve vs author `anthropic`; **FROZEN** — not rebased after attesting.

Closes #311


voodootikigod:
P5 evidence: cross-model review (codex), R1 code tree **approve, zero findings**; R2 final tree raised the expected post-attestation CRITICAL (repo-controlled approval in diff) with ZERO code findings — adjudicated by DEMONSTRATION per arc protocol: gate satisfied only with the correct key (no-key false, wrong-key false + untrustworthy chain, tampered in-diff file flips the digest), all probes reverted, tree byte-identical to attested. Live spot-check: Active listing went from 27 rows / 21 completed (78% wrong) to 6 rows / 0 completed, each of the six independently confirmed active. Tests written RED-first (5/5 failed on origin/main) incl. both boundary directions (completed+shipped stays stale; completed:false stays active). Mutation-gate 3/3 vs merge-base; preflight 54/54. Signed attestation openai vs anthropic @ git-change:ee8c87b9…:8d7db644…; gate GREEN in CI. FROZEN — do not rebase or Update-branch; bypass-merge.

PR #555, opened 2026-08-23
fix(core): --help prints synthesized usage for parseArgs configs without usage (#107)

### LINKED ISSUE #107: core parseArgs: --help still crashes for tools that pass no 'usage' (incomplete #72 fix)
## Context

Found during P5 prosecution on branch `cursor-native-flush`.

#76 (closing #72) added central `--help`/`-h` handling to `@adlc/core`'s `parseArgs` — but the handler only prints-and-exits when the calling tool passes a `usage` string/function (`packages/core/lib/cli.mjs`). When a tool passes no `usage`, `--help` falls through to Node's strict `parseArgs` and crashes with the exact `ERR_PARSE_ARGS_UNKNOWN_OPTION` stack trace #72 was filed about.

## Reproduction

```sh
node packages/rejection-mining/bin/rejection-mining.mjs --help
# TypeError [ERR_PARSE_ARGS_UNKNOWN_OPTION]: Unknown option '--help' ...
```

## Fix directions

1. In `parseArgs`, when `--help` is present and no `usage` is configured, print a generic fallback (e.g. derive a flags list from `config.options`) and exit 0 instead of falling through.
2. And/or audit which tool bins pass no `usage` (rejection-mining at minimum) and add usage strings.

## References

- `packages/core/lib/cli.mjs` (the conditional exit-0 path)
- `packages/rejection-mining/bin/rejection-mining.mjs` (no `usage` config)
- #72 / #76 (the original fix)

voodootikigod:
Ticket **T-01M0QGCCSWV8PXRB032WKG0T9B** (shard lands via #551).

## The bug

Seven of the 27 shipped bins crashed on `--help` and `-h` with a raw
`ERR_PARSE_ARGS_UNKNOWN_OPTION` stack trace and exit 1 — directly and through the
umbrella dispatcher (`adlc preflight --help`). All seven went through the single
`nodeParseArgs` call in `packages/core/lib/cli.mjs`: `parseArgs` printed-and-exited
only when `config.usage` was set, so usage-less configs fell through to `node:util`
parseArgs in strict mode, which throws on the undeclared `--help`. This is the first
command a new npm consumer runs.

## The fix

One file, `packages/core/lib/cli.mjs`: when a help flag is present and no `usage` is
configured, synthesize a listing from `config.options` (flag names, short aliases,
value placeholders, defaults), print it, exit 0. The `usage`-configured path is
untouched, and no bin was edited.

```
usage: preflight [options]

options:
  --test-cmd <value>
  --gh                (default: false)
  --llm               (default: false)
  --worktrees         (default: false)
  --json              (default: false)
  -h, --help          show this help
```

## Seven-bin sweep, and non-regression across all 31

Every bin under `packages/*/bin/*.mjs` was run with `--help` twice — once with the fix
stashed, once with it applied — and the outputs diffed:

| | before | after |
|---|---|---|
| behavior-diff, consensus-fix, lesson-foundry, model-router, preflight, rejection-mining, skill-rot | exit 1, stack trace | exit 0, synthesized listing |
| the other 24 bins | exit 0 | exit 0, **byte-identical** |

`diff -rq before/ after/` reports exactly those seven files as differing and no others.
`adlc preflight --help`, `adlc skill-rot --help` and `adlc behavior-diff --help` were
also confirmed through the dispatcher, each headed with the tool's own name.

## Test plan

- `packages/core/test/cli-help.test.mjs` (new, 14 tests). Written first: the 7
  new-behavior cases failed against unmodified core while the 7 non-regression cases
  already passed, so they pin current behavior rather than the change. Covers `--help`
  and `-h`, the exact rendered listing byte-for-byte, short aliases / `multiple` /
  defaults, an options-less config, the `argv[1]`-absent fallback, both configured-usage
  paths unchanged, configs declaring `help` or `h` themselves, and unknown flags still
  failing loudly. Program-name and exit-code claims run in spawned fixture bins, so
  nothing depends on what the test runner puts in `process.argv[1]`.
- `packages/cli/test/dispatch.test.mjs`: all seven bins asserted through `adlc <tool>
  --help` — exit 0, no stack trace, listing headed by the tool's name, not `adlc`.
  Verified load-bearing (fails on the merge-base `cli.mjs`, passes with the fix).
- Full suite: **54/54 segments green**.
- Mutation gate vs merge-base: **12/12 mutants killed**. An earlier round left the
  column-padding `off-by-one` alive, which is why the listing is now asserted byte-for-byte.

## Why `scripts/test/skills-catalog.test.mjs` changed

That test derives each tool's required arguments by probing it, and it took the first
usage-looking output as the contract. Once every tool answered `--help`, it read
`consensus-fix`'s generic flag listing, concluded "requires nothing", and **silently
retired its own coverage** — with every assertion still green. Two changes keep it honest:

1. The `--help` and bare-invocation probes are kept apart (and cached), so a tool that
   gains a generic `--help` cannot mask the `--test-cmd is required` its bare invocation states.
2. A synthesized listing is explicitly **not** read as a contract. Those tools stay in the
   file's known-gap list, unchanged — "we cannot check this" is not the same as "requires nothing".

Every pre-existing assertion in that file, including the known-gap allowlist, is
byte-identical to main. One test was added pinning the strength that was nearly lost;
disabling the detector fails both it and the allowlist assertion.

## Adversarial review

Five rounds, `--provider codex` (cross-model: reviewer openai vs author anthropic).
Read from `findings[]` and the model verdict, never the exit code.

| Round | Finding | Disposition |
|---|---|---|
| 1 | high — tests assume `process.argv[1]` under `node --test` | **Premise refuted** (13/13 passed under both `node --test <file>` and the repo runner, which shells out to `node --test`), **recommendation adopted**: program-name assertions moved into spawned fixtures. |
| 2, 4, 5 | medium — synthesized help omits required positionals/flags | **Accepted as a known limitation, out of scope.** `node:util` option specs carry no requiredness and no positional metadata, so core cannot know it; expressing it means editing the seven bins, which the ticket forbids. Strictly an improvement over a stack trace, and the repo's own tooling now encodes this exact fact. Follow-up: give those seven bins real `usage` strings. |
| 3 | medium — "violates the frozen-core contract" | **Refuted.** The rule (CONTRIBUTING.md:69, CONVENTIONS.md:35) is scope discipline for *tool package* PRs — "implement it locally in your `lib/`", "write ONLY inside your own `packages/<name>/`". Core is maintained through dedicated tickets under attestation; 44 commits on main touch `packages/core`, including #511 and #506. This ticket scopes the fix here deliberately. |
| 3 | medium — coverage records only consensus-fix's first failure | **Valid, pre-existing, now documented in place.** The bare probe sees only the first refusal because the tool exits at it, and feeding it arguments to uncover the next means running it for real — what `SAFE_BARE_PROBE` exists to prevent. The assertion is a floor, not a certificate; the catalog's taught invocation already passes `--files`. |
| 4 | medium — help name may come from the dispatcher | **Premise refuted** (the dispatcher spawns a child, which carries its own `argv[1]`), **coverage gap real and closed** by the new dispatcher test. |

| 6 | medium — cross-model approval bound to an older revision | **Valid and fixed.** The Node 18 portability commit changed the reviewed change set, so the first attestation no longer covered the branch. Re-recorded against the shipping revision; the manifest is append-only, so the superseded entry remains. |

Rounds 5, 6 and 7 verdict: **approve**. Round 7, on the exact attested tree, returned
only the known limitation above — five rounds converged on it with no new information.

## Node 18

CI's `test (18)` caught one case the local runtime did not: a fixture bin literally
named `.mjs` cannot be loaded by Node 18's ESM resolver. The branch under test is the
program-name derivation, not the loader, so the case now sets `process.argv[1]` instead.
Verified on Node 20 and 24 locally, and all 19 CI checks are green including
`test (18)`, `test (20)`, `test (22)` and the ubuntu/macos/windows ticket-store matrix.

## Attestation

`git-change:ee8c87b9aea4d9cf121a325a69c855e7dece180b:b4d86e05bd1b795e2225855a221731d5813a0d8f89a94b630fce397a00eefa2e`
— recorded openai-vs-anthropic, `tier-check --author-provider anthropic` PASS, and CI's
`gate` job green. (The superseded entry for `cc04a79…` stays in the append-only shard.)

Forge-resistance, each probe restored from git afterward: tamper (exit 2), strip
signature (2), zero signature (2), stale base (2), remove entry (2), attacker key
(2 — the recorder itself refuses, "chain broken at seq 96: signature invalid").

## Scope note

The ticket scopes `packages/core/lib/cli.mjs` and `packages/core/test/**`. Two files
outside it were touched, both consequences of the fix rather than additions to it:
`scripts/test/skills-catalog.test.mjs` (would otherwise fail outright and shed coverage,
see above) and `packages/cli/test/dispatch.test.mjs` (the dispatcher path the issue names).

Closes #107


voodootikigod:
P5 evidence: 7 codex review rounds on the way to a final-round **approve on the exact attested tree** (findings[] + verdict read per arc rule). Real findings fixed or adopted: program-name assertions moved into spawned fixtures; dispatcher-path coverage added; a stale-revision approve caught and re-recorded (append-only shard carries the superseded entry — correct). Refutations evidenced, incl. the frozen-core claim (44 core commits on main; scope discipline misread). Known limitation, five rounds converged, out of scope by ticket design: synthesized help cannot express required positionals (node:util specs carry no requiredness) — follow-up filed to give the seven bins real usage strings. Seven-bin sweep proven BOTH directions with byte-diff (exactly the seven change; 24 others identical; dispatcher path confirmed per-bin). Deliberate scope expansion, approved: scripts/test/skills-catalog.test.mjs would have SILENTLY RETIRED its own coverage once tools answered --help (synthesized listing read as 'requires nothing') — the detector now keeps those tools in the known-gap list, with a test proving the pin load-bearing. Mutation-gate 12/12; suite 54/54; forge-resistance six-way incl. attacker-key refused at record time. Signed attestation openai vs anthropic @ git-change:ee8c87b9…:b4d86e05…; gate GREEN. FROZEN at e9d5ee63 — do not rebase or Update-branch; bypass-merge.

PR #556, opened 2026-08-23
fix(cli): resolve npx beside the node executable so `adlc review` is reachable on Windows (#233)

### LINKED ISSUE #233: adlc review is unreachable on Windows: runExternal spawns bare npx (ENOENT), and npx.cmd alone gives EINVAL
`adlc review` cannot run on Windows at all. It fails before reaching `adversarial-review`, so on Windows the P5 external-review verb is effectively absent — and it fails in a way that reads like a missing package rather than a spawn bug, which is how we ran for two weeks believing the reviewer was simply not installed.

## Repro

```
> adlc review --help
error: failed to run npx adversarial-review: spawnSync npx ENOENT
```

`adversarial-review` is installed globally (2.8.0) and `npx` resolves fine in the shell.

## Cause

`lib/dispatch.mjs`:

```js
function runExternal(packageName, args, spawnFn) {
  const result = spawnFn('npx', [packageName, ...args], { stdio: 'inherit' });
```

On Windows the executable is `npx.cmd`, and Node does **not** do PATHEXT resolution for a non-shell spawn — so bare `'npx'` is ENOENT.

## The part worth knowing: the obvious half-fix makes it worse

Changing `'npx'` → `'npx.cmd'` alone moves the error, it does not fix it:

```
error: failed to run npx adversarial-review: spawnSync npx.cmd EINVAL
```

Node ≥18.20.2 / 20.12.2 / 21.7.3 **refuses to spawn `.bat`/`.cmd` without `shell: true`** — that is the CVE-2024-27980 argument-injection mitigation. So both halves are required.

## What worked

```js
const win = process.platform === 'win32';
const result = spawnFn(win ? 'npx.cmd' : 'npx', [packageName, ...args],
                       { stdio: 'inherit', shell: win });
```

Scoping `shell` to win32 rather than setting it unconditionally keeps POSIX argument quoting unchanged.

Verified against `@adlc/cli@1.5.1` on Windows 11, Node via scoop: `adlc review --base main --prompt-only` now emits the full adversarial prompt.

## Note on `shell: true` and quoting

If you take this, the `shell: true` path means `args` are re-parsed by cmd.exe. Passthrough args here come from adlc's own CLI surface, but a focus string containing `&` or `|` (e.g. `adlc review "check auth & session paths"`) would be split by the shell. Quoting each arg before the spawn, or preferring `execFile`-style resolution of the real `npx.cmd` path, would be more robust than my minimal patch.

## Suggested test

A unit test asserting `runExternal` is invoked with a shell-capable spawn on `win32` would pin this — the current failure is invisible on CI if CI is Linux-only.

I am carrying this as a local re-appliable patch script in the meantime, since a global `npm i -g` wipes it. Happy to open a PR if the shape above looks right to you.

voodootikigod:
Ticket: **T-01M0QGCDC87DTK2NRPY9A3R1G0** (ticket shard lands via #551, still open — attestation recorded with `--allow-unknown-ticket`)

Closes #233

## The bug

`runExternal` (packages/cli/lib/dispatch.mjs) spawned the bare name `npx` with `{stdio:'inherit'}` and no shell. Node does no PATHEXT resolution for a non-shell spawn, so on win32 `npx` — a name with no extension — is ENOENT. `review` is the only `external: true` verb, so `adlc review` never reached the `adversarial-review` CLI at all, and the failure read like a missing package.

## Chosen mechanism, and why

Resolve the `npx-cli.js` that npm ships with the **running** node and spawn `node <npx-cli.js> <package> ...args`.

This is not a workaround — it is exactly what npm's own shims do on their last line. Both end identically:

```
npx.cmd   SET "NPX_CLI_JS=%~dp0\node_modules\npm\bin\npx-cli.js"
          "%NODE_EXE%" "%NPX_CLI_JS%" %*

npx       NPX_CLI_JS="$CLI_BASEDIR/node_modules/npm/bin/npx-cli.js"
          "$NODE_EXE" "$NPX_CLI_JS" "$@"
```

`%~dp0` is the install dir the shim shares with `node.exe`, and `CLI_BASEDIR` is literally `dirname(process.execPath)`. So this is the shims' behaviour minus cmd.exe, and it needs **no shell on any platform**.

### Why not `npx.cmd` + `shell: true` (CVE-2024-27980)

The obvious repair — spawn the sibling `npx.cmd` — forces a shell. Since the CVE-2024-27980 hardening (Node ≥18.20.2 / 20.12.2 / 21.7.3) spawning a `.cmd` **without** `shell: true` throws `ERR_CHILD_PROCESS_UNSUPPORTED_SPAWN`; that hardening exists precisely because `.cmd` files are interpreted by `cmd.exe`, which re-parses the command line. Turning the shell on would hand cmd.exe every `adlc review` argument — user-supplied text like `--focus 'a & b | c'` — to re-parse as operators.

**No path in this change requests a shell**, and argv is passed as an array throughout. Both properties are pinned by tests (`no dispatch path ever asks for a shell` asserts `options.shell === undefined` across all six resolution outcomes; `shell metacharacters in user args survive as literal separate argv entries` pins `&`, `|`, `^`, `"`, `>` arriving verbatim). `npx.cmd` is therefore never spawned, and the win32 branch degrades rather than reaching for a shell.

### Evidence for the Windows layout

Read out of npm's own shipped shims rather than from memory. `npx.cmd` asserts `SET "NODE_EXE=%~dp0\node.exe"` — npm itself expects `node.exe` and `node_modules\npm\bin\npx-cli.js` to share `%~dp0`. The POSIX `<prefix>/bin/node` + `<prefix>/lib/node_modules/npm/bin/npx-cli.js` layout was verified directly on this machine (the `bin/npx` symlink resolves exactly there).

### Resolution ladder

1. `<execDir>/node_modules/npm/bin/npx-cli.js` → `node <cli> <pkg> …` (win32 + posix)
2. `<execDir>/../lib/node_modules/npm/bin/npx-cli.js` → same (**posix only**)
3. posix only: the sibling `<execDir>/npx`, exec'd directly
4. bare `npx` — the pre-existing behaviour, unchanged

The table is **platform-keyed on purpose**. Rung 2 walks `..` out of the install dir; on POSIX that stays inside the install prefix, but on Windows it lands on a sibling npm never creates and an unprivileged user often can — an nvm-windows node at `...\nvm\v20.11.0\node.exe` would probe `...\nvm\lib\node_modules\npm\bin\npx-cli.js`. Anything found there would be run as `node <file>`, so win32 probes the install dir and nothing else.

### Candidate validation

Each candidate is validated for the way it is actually invoked, and a rejected candidate **falls through to the next rung** rather than ending the search:

- **non-empty regular file** for `npx-cli.js`. `existsSync` is true for directories; `node <directory>` exits non-zero, but `node <empty file>` exits **zero**, which `runChild` propagates as success. An interrupted npm install would otherwise silently turn the review gate green with no reviewer having run.
- **readable** (`R_OK`). `statSync` succeeds on a file this process cannot read; without the check, an unreadable candidate in rung 1 would be selected and die with EACCES while a good npm sat in rung 2.
- **executable** (`X_OK`) additionally for the sibling, which the OS execs itself rather than handing to node.

An unresolved npx now names the paths it probed, so the failure is distinguishable from a missing package.

## Security posture

Every probed path derives from `process.execPath` and literal constants — no user or env input reaches it, and PATH is deliberately **not** consulted to derive an executable (that is the trust surface this suite is moving away from). The `npxEnv` injection seam is unreachable from the CLI: `packages/cli/bin/adlc.mjs:53` calls `dispatch(first, argv.slice(1))` with no opts, and `dispatch` has one non-test call site repo-wide.

## Test plan

**No Windows job in CI covers `@adlc/cli`** — `.github/workflows/ticket-store-platform.yml` is the only Windows matrix and runs `npm test --workspace @adlc/tickets` only. The platform-branch tests are therefore the entire regression net for win32. They inject `platform` / `execPath` / `isFile` / `isExecutable` and assert the exact spawn argv from a Linux runner.

- [x] `node --test packages/cli/test/*.test.mjs` → **53 passing, 0 failing, 0 skipped**
- [x] `node scripts/mutation-gate.mjs <merge-base>` → **12/12 mutants killed, exit 0**
- [x] Hand-planted mutants, all killed: layout order swapped; layout table reversed; a wider `../../` entry prepended; win32 probing the posix table; `isFile` weakened to `existsSync`; a non-file candidate aborting the ladder; the unresolved diagnostic dropped; `shell: true` added; `stdio: 'pipe'`; argv reordered.
- [x] Live end-to-end on this machine: `dispatch('review', ['--help'])` resolves `node <…>/lib/node_modules/npm/bin/npx-cli.js` and prints adversarial-review's own help, proving the resolved invocation reaches the real CLI.
- [x] Real-filesystem tests (not just injected fakes) for the directory, zero-byte, unreadable, and non-executable cases. The readability test skips under uid 0, where permission bits do not restrain the caller and the assertion would silently invert.

Non-Windows behaviour is unchanged in effect: it resolves the same npx program it resolved before (verified: `which npx` → the same `npx-cli.js` this returns). The spawn *shape* changes from `spawn('npx', [pkg, …])` to `spawn(node, [npxCli, pkg, …])`, so `review-dispatch.test.mjs` was updated to pin the new literal argv.

## Review

Six adversarial-review rounds (`--provider codex`, distinct from the authoring family) plus an in-house P5 prosecutor pass in a separate worktree. Findings fixed across the loop: unpinned layout ordering and the win32 install-dir escape; `existsSync` accepting directories; zero-byte, unreadable and non-executable candidates; a missing probe diagnostic; and a containment assertion that used mixed path separators and would have failed on Windows. Final round: **approve, 0 findings**, 3/3 files examined, none skipped.

Two findings were adjudicated as **not merge-blocking**, with reasoning:

- *A truncated-but-non-empty `npx-cli.js` can still exit 0 silently.* Not a regression — `main` has the identical exposure today, since the bare `npx` it spawns resolves through the same shim ending in `node <npx-cli.js>`. The recommended repair (sniffing the file for an npm entrypoint) is brittle across npm versions and does not close the class. The real defect is that a review gate treats "child exited 0" as success at all, which is a wider change to `runChild` affecting every verb.
- *Windows installs with npm under a separate prefix still fall through to bare `npx`.* Also not a regression — that configuration fails today and fails identically after. The suggested repair reintroduces PATH-derived executable resolution, which the ticket's design constraint explicitly names as the direction being moved away from. The failure is now diagnosable rather than silent.

Both are follow-up candidates rather than fixes for this ticket.


voodootikigod:
P5 evidence: 6 codex review rounds + an independent in-house prosecution (own worktree, 21 planted mutants) — the lane REJECTED a clean round-1 approve on a trust-root file and prosecuted anyway, which surfaced 3 survivors incl. the unpinned layout table and a directory-accepted-as-file hole; rounds 2-5 fixed four real findings (empty-file, execute-bit X_OK, unreadable-candidate R_OK fallthrough, and a mixed-separator containment assertion that would have failed on the one platform this fix protects); round 6 **clean approve, zero findings**. Mechanism proven from npm's own shims (both end in `node <npx-cli.js>`), no shell anywhere, argv untouched, rung-2 platform-gated against the nvm-windows escape. Two findings refuted with evidence and filed as follow-ups (pre-existing exposure, not regressions). Mutation-gate 12/12 vs merge-base + 10 hand-planted mutants all killed. Signed attestation openai vs anthropic @ git-change:ee8c87b9…:c5ba9475…; gate GREEN; revision unchanged post-shard. FROZEN at d2cf00ab — do not rebase or Update-branch; bypass-merge. CI note: no Windows job exercises @adlc/cli — the injected platform-branch tests are the entire win32 regression net.

PR #565, opened 2026-08-24
fix(claude-code): an import-time failure in the wrapper's bin detection denies, never crashes fail-open

voodootikigod:
Ticket: **T-01M0122Z30GW4GXEYW948FMYXF** — *CC wrapper: an import-time failure in bin detection must deny (exit 2), never crash fail-open*

Source: agy cross-model review of PR #497 (seam 6), round 1. The ticket required verifying the premise before building — two of three agy findings on 2026-08-14 were refuted against the real code. This one reproduces.

## The reproduced defect

`plugins/adlc-claude-code/hooks/adlc-hook-run.mjs:102-104` resolved its own entry point with an unguarded `realpathSync(process.argv[1])` during module evaluation. On `origin/main`:

```
$ node --input-type=module -e "process.argv[1]='/nonexistent/virtual-entry.mjs'; \
    await import('file://.../hooks/adlc-hook-run.mjs')"
Error: ENOENT: no such file or directory, lstat '/nonexistent'
    at realpathSync (node:fs:2775:29)
    at .../adlc-hook-run.mjs:103:71
exit=1
```

A *directory* `argv[1]` resolves fine (exit 0) — only an unresolvable path throws, which is what a bundler, an embedded REPL, or a test runner with a synthetic entry point hands you.

Why it matters here specifically: seam 6 (#497) established that in Claude Code **only exit 2 blocks a tool call**; exit 1 is a non-blocking error that lets the tool proceed. This crash exits 1 *before* `dispatch()` and its timeout handling ever run, so it bypasses the seam-6 enforcing-deny entirely — the exact fail-open path that seam existed to close.

## Fix semantics, per mode

`entryPointState()` is now three-valued and never throws:

| state | meaning | enforcing (`rails`/`buildgate`/`handoff`) | advisory |
|---|---|---|---|
| `yes` | argv[1] is this module | dispatch, unchanged | dispatch, unchanged |
| `no` | argv[1] resolves to something else | quiet exit 0, unchanged | quiet exit 0, unchanged |
| `unknown` | argv[1] unresolvable | **exit 2 + deny message on stderr** | diagnostic on stderr, exit 0 |

`unknown` deliberately does not collapse into `no`. Declining to dispatch also exits 0, which lets the tool through ungated — the same fail-open hole from the other direction. When the wrapper cannot prove it is *not* the entry point, an enforcing hook denies.

A plain string match now precedes any filesystem probe, so a broken filesystem cannot demote a genuine direct invocation into a deny; the `realpath` calls exist only to see through symlinks (the ESM loader canonicalises `import.meta.url`, node does not canonicalise `argv[1]`).

A top-level `try/catch` around module evaluation applies the same fail-closed direction to any other error the wrapper raises before dispatching.

Advisory modes do not start blocking anywhere: they get the diagnostic, exit 0, and the message is never worded as a block.

## Test plan

- `plugins/adlc-claude-code/hooks/test/wrapper-import-time-deny.test.mjs` — spawn-based, asserting real exit codes: every enforcing mode denies with 2 and a deny message naming the unresolvable input; no `ENOENT`, no stack frames; every advisory mode stays at 0; a resolvable non-wrapper `argv[1]` and an absent `argv[1]` import cleanly.
  - This file **deliberately does not import the wrapper**. Evaluating that module can exit the process by design, so an entry-point defect would kill an importing test file mid-load — silently, with status 0. The mutation gate found exactly that: four survivors were mutants that destroyed the suite before it could run. A `IMPORT-COMPLETED` sentinel on stdout now proves module evaluation handed control back rather than dispatching.
- `plugins/adlc-claude-code/hooks/test/wrapper-entry-point.test.mjs` — `entryPointState` with an injected `realpath` (exact match takes no fs probe, symlink, different file, absent, unresolvable on either side, never throws) plus a real on-disk symlink, and `importFailureExitCode` in both directions.
- AC2: the seam-6 suite is unmodified and green — 242 pre-existing tests, 278 total with the 36 new ones (`node --test plugins/adlc-claude-code/hooks/test/*.test.mjs`).
- AC3: `npm test` — 54/54 segments.
- Mutation gate vs merge-base `d7cd871`: **20/20 mutants killed, 0 survived.**

---
🤖 Session: https://claude.ai/code/session_011Fbv2GiKerPz67RapZigRM


voodootikigod:
## Cross-model prosecution record

`npx adversarial-review --base origin/main --provider codex --timeout 600 --json`, three rounds. Findings and the model verdict were read directly — not exit codes.

**Round 1 — 1 HIGH, `test-integrity`/`security`: "Dispatch startup failures still allow enforcing tool calls."**
Verified as **real but pre-existing and deliberately out of scope** → filed as #567 rather than fixed here.
- `dispatch()` is byte-identical to `main` in this PR; only its call site changed.
- AC2 of the ticket requires the seam-6 suite stay green *unmodified*, and two of its tests explicitly pin the current contract (`'non-timeout spawn error → operational error (1)'`, `'missing hook script → operational error (1)'`). Fixing it here means rewriting those assertions.
- The fix carries a real tradeoff needing a deliberate decision: today a broken install degrades to "hooks don't run"; denying turns it into "you cannot edit any file until the install is repaired" — and the repair is blocked too. That escape hatch has to be designed alongside, not flipped in passing.
- The asymmetry is now recorded at the code site so it does not read as an oversight.

Note: the harness halved this finding's confidence because the quoted evidence spliced two separate regions and did not match the context verbatim. It was **not** dismissed on that technicality — the claim was checked against the file independently and is accurate.

**Round 2 — verdict approve, 1 LOW `test-integrity`: temp-dir leak in the on-disk symlink fixture.** Valid; fixed with `rmSync` in a `finally`.

**Round 3 — verdict approve, `findings: []`.** Clean round on the exact tree attested.

## Gates

- Mutation gate vs merge-base `d7cd8710`: **20/20 killed, 0 survived**. Four of the eight round-1 survivors were mutants that made the wrapper self-dispatch at import and *destroy the importing test file before it ran* (exit 0, no assertions). That is why the spawn-based tests were split into a file that never imports the module, with a stdout sentinel proving module evaluation handed control back.
- `node --test plugins/adlc-claude-code/hooks/test/*.test.mjs` — 278 pass (242 pre-existing seam-6 tests unmodified + 36 new).
- `npm test` — 54/54 segments.
- Trust-root tier (rails deny-path of T37, `plugins/adlc-claude-code/**`): cross-model approve recorded, `openai` reviewer / `anthropic` author, signed (`sigVersion` 2, seq 271), bound to revision `git-change:d7cd8710:cc824ab3`. `tier-check` reports `satisfied: true`, `chainTrustworthy: true`, `truncationDetected: false`.

voodootikigod:
P5 evidence: cross-model review (codex), 3 rounds → R3 clean (findings []); no pre-approval CRITICAL (gate went green first push post-attest). Premise CONFIRMED and fixed: entryPointState() is now three-valued and never throws — the 'unknown' (unresolvable argv[1]) case denies exit 2 on enforcing modes / diagnostic+exit-0 on advisory, deliberately NOT collapsing into 'no' (which would fail open from the other direction); a string match precedes any fs probe so a broken FS can't demote a real direct invocation; top-level try/catch applies the same fail-closed direction to any module-eval error. R1 HIGH (dispatch startup exit-1 returns still allow enforcing calls) verified real but PRE-EXISTING + out of scope (dispatch() byte-identical to main; AC2 requires the seam-6 suite unmodified and two of its tests pin those returns; genuine repair-vs-deny tradeoff) → filed #567, asymmetry recorded at the code. Mutation-gate 20/20 vs merge-base; CC suite 278 (242 seam-6 unmodified + 36 new); 54/54 suite. Signed attestation openai vs anthropic @ git-change:d7cd8710…:cc824ab3…, seq 271, gate GREEN. FROZEN at 300b1641 — do not rebase or Update-branch.

Issue #104, opened 2026-07-05
ticket-prune --write output is CI-unmergeable: rails-guard-ci denies base-ticket removal


voodootikigod:
## Context

Found dogfooding both tools together on the cursor-native-parity build (P1 design review caught it before the branch reached CI).

`adlc ticket-prune --write` (#39/#91) archives stale tickets **out of** `.adlc/tickets.json` (into gitignored `tickets.archive.json`). But `scripts/rails-guard-ci.mjs`'s `assertBaseTicketContractsPreserved` **hard-denies removing any base-ref ticket from `tickets.json` in a PR** (exit 2: `base ticket T1 cannot be removed from .adlc/tickets.json in a PR`), and `.github/workflows/ci.yml` runs that gate on every PR.

Net effect: the prune tool's entire write path produces a diff that can never merge through CI. The two shipped features are in direct conflict — pruning is only possible via the protected-base admin path, which no doc mentions.

## Reproduction

```sh
node packages/ticket-prune/bin/ticket-prune.mjs --write   # archives stale tickets
node scripts/rails-guard-ci.mjs origin/main                # exit 2, base ticket cannot be removed
```

## Possible directions (not mutually exclusive)

1. **Teach the CI gate about archival**: allow removal of a base ticket from `tickets.json` iff the same PR's `tickets.archive.json`… is gitignored, so that evidence isn't visible to CI. Would require tracking the archive file, or an in-`tickets.json` tombstone (`status: archived`) instead of physical removal.
2. **Change ticket-prune to tombstone instead of remove**: `--write` sets `status: "archived"` in place (append-only-friendly, CI-safe), with actual removal reserved for the protected-base ceremony.
3. **Document the constraint**: at minimum, `docs/tools/ticket-prune.md` + the command docs must say the write path requires the protected-base runner ceremony / admin merge, not a normal PR.

Option 2 feels most aligned with the trust-root doctrine (the base contract only ever grows or gets annotated in PRs; destructive edits need the ceremony).

## References

- `packages/ticket-prune/lib/run.mjs` (#91)
- `scripts/rails-guard-ci.mjs` `assertBaseTicketContractsPreserved`
- Caught by the P1 adversarial design review on the cursor-native-parity build (branch `cursor-native-flush`)

PR #571, opened 2026-08-24
fix(rails-guard): anchor the ticket-store add-vs-alter comparison to the merge-base (#506 pattern)

voodootikigod:
Ticket **T-01M0122YKBY211SPJNFGPCH4TN** — rails-guard-ci: anchor the ticket-store add-vs-alter comparison to the merge-base, not the base tip.

## The false red this fixes

`assertBaseTicketContractsPreserved` compared the ticket store at the **base TIP** against the PR's HEAD tree. Any store change the base absorbed *after* the branch point therefore entered the comparison as a reverse-direction difference on tickets the PR never touched.

Reproduced on 2026-08-14: PR #509 batch-marked 42 tickets `completed: true` on main. PR #493 — nine added shards, no ticket touched — was then denied with

```
base ticket T-01KYJ3NMV8ZZDFD9DB31DJEGJF contract cannot change in a PR
```

because its stale tree still carried that shard *without* the completed flag. A ticket **added** on the base reads as a **removal** the same way. The only workaround was a rebase, and every long-lived PR re-hits it after the next store change on base.

This is the identical stale-base defect class fixed for trust-root tiering and change-set identity in #506, one gate over, and it takes the same fix.

## Anchoring semantics

The comparison basis becomes the store at `git merge-base --all <trustedBase> HEAD`, exactly-one-or-fail-closed (ambiguous criss-cross, disjoint histories, and an unresolvable HEAD all fail closed to exit 1 — never a fall-back to the tip, which is the basis being replaced).

Merge semantics agree: a ticket the PR leaves at its merge-base content contributes nothing when the change merges (three-way merge resolves to the base side), and a ticket added only on the base is kept by the merge, never deleted by it.

**Deliberately not re-anchored**, each pinned by a test:

- the **rails union** and the **T36 `completed` trust anchor** still read the BASE TIP. Reading completion from the tip is the documented forge-resistance design; reading rails from the tip can only *widen* the frozen set for a stale PR — the fail-safe direction.
- the store-level "base ticket store was removed or renamed at HEAD" denial stays tip-anchored.
- the migration ceremony (format-version and logical-hash comparison) is untouched; the merge-base is resolved lazily inside the non-migration branch, so a migration PR gains no new failure mode.

A store **absent** at the merge-base (a branch predating the store) is a vacuous comparison rather than a mass phantom removal; any *other* load error fails closed exactly as the base-tip load does.

## Cross-model review found a real hole in the first draft

Round 1 (codex) raised a HIGH that was correct and is fixed here. Merge-base anchoring alone widened the bounded #104/T36 completion allowance: a ticket railless at the branch point that **gains rails on the base afterward** could be annotated `completed: true` by a stale PR. I measured the merge — git combines a `rails` array inserted near the top of a shard with a `completed: true` appended at the bottom **without a conflict** — so the merged ticket carries both, and T36 reads completion from the tip, expiring a rail no ceremony ever lifted.

The draft's justifying comment (that a conflict-free merge implies the base never touched the same ticket) was simply false; it is corrected in place rather than deleted. `isCompletionAnnotationOnly` now requires railless at **both** anchors, which only ever denies more than the merge-base check alone and never more than the base-tip check it replaced. Its `railedAtBaseTip` parameter defaults to the conservative value so an omission fails closed.

## Tests

18 tests in `packages/rails-guard/test/ticket-store-merge-base.test.mjs`, on a fixture repo whose base tip advances **independently** of the PR branch — a fixture where the two anchors coincide cannot tell them apart. All 18 fail against the pre-change gate.

- AC1 — the #493/#509 false red in both directions (base-side completion, base-side add)
- AC2/AC3 — genuine alteration and removal still denied, and denied on the *correct* ticket
- AC4 — the bounded completion allowance still allowed, and still bounded to railless tickets
- R1 regression — rails added on the base after the branch point revoke the allowance (fixture + unit)
- AC5 — the tip anchor pinned in both directions (a rail expired at the tip is expired here; a rail declared at the tip still binds a stale PR)
- the directory-store denial names the directory store, not `.adlc/tickets.json`
- absent vs unreadable merge-base store; disjoint and criss-cross histories fail closed
- `resolveMergeBase`'s HEAD guard through the module's own git seam (real git moves status and stdout together; only a stub separates them)

`node --test packages/rails-guard/test/*.test.mjs scripts/test/rails-guard*.test.mjs` → 557/557. Full `npm test` → 54/54 segments. Mutation gate vs merge-base → 12/12 killed, 0 survived.

## Expected CI reds

1. **rails-guard ceremony red (#141-style), expected.** This edits `packages/rails-guard/lib/ci/**`, an immutable trust root. Merging needs the `trust-root-change` label plus a non-author CODEOWNER approval, or the protected-base admin path.
2. **Cross-model gate.** Enforcement-package change → trust-root tier. Attestation recorded: reviewer `openai` (codex CLI) vs author `anthropic`, bound to revision `git-change:d7cd8710:01adad05`, signed (`sigVersion: 2`). `tier-check --author-provider anthropic` → `satisfied: true`, `chainTrustworthy: true`, `truncationDetected: false`.

## Review rounds

| Round | Findings | Outcome |
|---|---|---|
| R1 | 1 HIGH — completion allowance widened by merge-base anchoring | Real. Confirmed by measuring the merge. Fixed + regression test. |
| R2 | 1 MEDIUM — merge-base store read bypasses the gate's git timeout/buffer | Refuted (below). |
| R3 | none | Clean on the code. |
| R4 | 1 CRITICAL — "repository-controlled manifest claims approval" | The expected pre-approval artifact; forge-resistance table below. |

**R2 refuted with evidence.** The gate already loads the base-**tip** store through the identical unbounded `GitTreeTicketStore` at `rail-freeze.mjs:86`, on every run; `rail-freeze.mjs:379` is the same construct against a different revision. The 64 MiB / no-timeout bounds live in `packages/tickets/lib/stores/git-tree.mjs`, which this PR does not touch and which is outside the ticket's `packages/rails-guard/**` scope. The failure direction is exit 1 — fail closed. Bounding *both* store reads is a reasonable follow-up ticket against `packages/tickets`; doing it here would be an out-of-scope edit to a second package.

**R4 — forge-resistance table.** The finding fires on the presence of the attestation record in the diff. That record is evidence to be *verified*, never a claim to be believed, and the reviewer is right not to credit it:

| Property | Mechanism | Verified |
|---|---|---|
| Revision-bound | identity is `git-change:<merge-base>:<content digest>`, recomputed from the working tree at check time | Appending one comment byte to `git.mjs` flipped the digest to `c4522149…` and `satisfied` to **false** |
| Signature-bound | HMAC `sigVersion: 2`; `hasCrossModelApproveForRevision` returns false without the key and on an untrustworthy chain | Editing the recorded verdict text in place gave `chainTrustworthy: false`, `satisfied: false` |
| Distinct-provider | `record-cross-model` fails closed when `--provider === --author-provider`; verification is author-scoped | reviewer `openai` vs author `anthropic` |
| Key not in the tree | `ADLC_MANIFEST_KEY` is a repo secret; the verifying job holds it via `pull_request_target` | a PR branch cannot mint a valid signature |
| Append-only | segment anchored at `root` seq 271 with a line hash; rail-freeze §9.2 validates segments append-only | — |

Completion follows separately. Do not merge without the trust-root ceremony.


voodootikigod:
P5 evidence: cross-model review (codex) → R3 clean (findings []), R4 post-attest pre-approval CRITICAL adjudicated by a forge-resistance table with two rows proven EMPIRICALLY (a one-byte comment append flipped the revision digest + satisfied→false; editing the recorded verdict gave chainTrustworthy→false; both reverted). R1 caught a real HIGH in the first draft — merge-base anchoring ALONE widens the T36 completion allowance (a stale PR could annotate completed:true on a ticket that gained rails on the base side), PROVEN by an actual conflict-free git merge, not argued; fix requires railless at BOTH anchors, param defaulting conservative (fail-closed). R2 MEDIUM refuted with evidence (the unbounded store read pre-exists at rail-freeze.mjs:86, outside scope). 557/557 package tests; 54/54 suite; mutation-gate 12/12. Signed attestation openai vs anthropic @ git-change:d7cd8710…:01adad05…, anchored root seq 271. Two EXPECTED reds: rails-guard ceremony (edits packages/rails-guard/lib/ci/** trust root — needs the trust-root-change label + non-author CODEOWNER approval) and the cross-model gate (satisfied by the attestation). FROZEN at 2ed2a06b — do not rebase or Update-branch.

PR #575, opened 2026-08-24
chore: repo hygiene before the 1.11.0 release

voodootikigod:
## Why

Prep pass for the 1.11.0 release. The suite has grown from 23 to **38 publish
targets** (31 under `packages/*`, 7 under `plugins/*`) without the front-door
docs following, and some scratch had accumulated at the repo root.

Docs and comments only — no behavior change.

## Every shipped artifact is now discoverable

`docs/package-reference.md` listed **25 of 31** `packages/*` and **1 of 7**
harness plugin packages. Twelve public, published npm packages were absent from
the project's own reference — installable, but not findable from our docs:

| Added to the packages table | Added to the harness-plugin list |
| --- | --- |
| `@adlc/context-handoff`, `@adlc/fleet`, `@adlc/init`, `@adlc/quartermaster`, `@adlc/ticket-sync`, `@adlc/tickets` | `@adlc/codex`, `@adlc/cursor`, `@adlc/gemini`, `@adlc/herdr`, `@adlc/opencode`, `@adlc/pi` |

- README's toolkit table gained the same six `packages/*` entries.
- The **Package groups** section now mirrors `packages/cli/lib/registry.mjs` —
  the authoritative list behind `adlc --help`, which had *not* drifted. The
  registry was right; the prose was behind it.
- `@adlc/quartermaster` and `@adlc/ticket-sync` are called out as the two
  targets with no `adlc` subcommand (a library and a standalone binary), so the
  gap between the package set and the verb set is documented rather than
  looking like an omission.

## Release doc that could mislead an operator mid-release

`docs/RELEASING.md` asserted "all 23 packages" in two places, including the
step describing what the publish workflow pushes. Rewritten to describe the
publish set **by rule** ("every non-private package under `packages/*` and
`plugins/*`", pointing at the `verify` sweep in `.claude/release-profile.md`)
rather than by count, so it cannot go stale on the next package added.

## Root cleanup

- **Removed `0`** — an empty file committed by accident in #116.
- **`delta.md` → `docs/reviews/`**, alongside the other audit reports. The four
  test comments citing it as provenance (`Seam 6/7 remediation (delta.md / PR
  #486)`) are repointed to the new path, so the citations stay accurate.
- **Dropped the two unreferenced ADLC blog drafts** (`adlc-9-…`, `adlc-10-…`).
  They had zero inbound references and now live with the rest of the series in
  the site repo.

The root is now only standard project files.

## Deliberately not touched

[...]

PR #542, opened 2026-08-22
fix(rails-guard): remove package.json/.npmrc from default immutable trust roots

voodootikigod:
Ticket: **T-01M0MX7YX0JJQS7M8S1SGG5BJS** (tickets land via #538)

## Blocker

Release-audit found an undeclared breaking change in `pkg:rails-guard`. PR #507 added
`package.json` and `.npmrc` to `DEFAULT_IMMUTABLE_TRUST_ROOTS`. That set is not opt-out —
`resolveImmutableTrustRoots({ additional })` only ever concatenates — so as of the 1.11.0
window every consuming repo has two files it edits constantly frozen behind the #141
ceremony, and no way to decline.

The only escape, `isVersionOnlyChange`, covers version fields and `@adlc/*` dependency
repins on manifest basenames. A third-party dependency bump on the root manifest is not
covered, and `.npmrc` has no exemption at all. Audit reproduction: a consumer-shaped
fixture with a root `package.json` devDependency bump exits **2** on HEAD and **0** on
v1.10.0 — same input, opposite verdict, minor bump, no CHANGELOG note, no README mention.

Decision (Chris, 2026-08-22): audit option (a) — move the two paths out of the frozen
defaults and keep this repo's protection through the existing `--trust-root` mechanism.

## What changed

- `packages/rails-guard/lib/ci/trust-roots.mjs` — `package.json` and `.npmrc` (and their
  rationale comment) removed from the frozen defaults. The gate-source rationale header is
  untouched: it is cross-model-reviewed (#363) and still authoritative for the remaining
  entries. `resolveImmutableTrustRoots` semantics are unchanged.
- `scripts/rails-guard-ci.mjs` — both paths added to `REPO_TRUST_ROOTS`, which the wrapper
  already flat-maps into `--trust-root` flags. **This repo's posture is unchanged.**
- `packages/rails-guard/test/ci-trust-roots.test.mjs` — the default-set expectation is
  updated and pinned in the negative direction (a test now fails if either path returns to
  the defaults). The `package.json` ceremony scenarios are repointed through the
  `--trust-root` route — the real ADLC-repo posture — and still deny. New consumer
  regression: a repo that did not opt in gets a clean pass on a devDependency bump, and the
  identical diff denies once `package.json` is passed as a trust root, so the opt-in is
  shown to still buy something.
- `packages/rails-guard/README.md` — new "Immutable trust roots (CI gate)" section: what
  the frozen defaults protect, that these two paths are deliberately not defaults, how
[...]

PR #576, opened 2026-08-24
chore: bump version to 1.11.0

voodootikigod:
Lockstep version bump for the **1.11.0** release. `1.10.0 → 1.11.0`.

Generated by `node scripts/release.mjs 1.11.0` (no `--publish`) plus
`node scripts/changelog.mjs 1.11.0`, then hand-curated. No hand-edited manifests.

## Contents

- **47 manifests + the lockfile** moved to 1.11.0 — `@adlc/core`, the phase CLIs,
  the `@adlc/cli` umbrella, every versioned `plugins/*` package, the root, and both
  marketplace manifests. Internal `@adlc/*` deps repinned by the script, range style
  preserved.
- **`CHANGELOG.md`** — `[1.11.0]` curated.

## Changelog curation

`scripts/changelog.mjs` derives a section from the 72 commits since v1.10.0, but it
appends a *new* `[1.11.0]` above the existing `[Unreleased]` rather than folding that
hand-written content in — so the two curated entries (the rails-guard trust-root
change and the build-gate/Codex `countToolCalls` fix) were left stranded below, and
the release would have shipped terse commit-subject versions of both instead.

Curation applied:

- Folded both hand-written entries into `[1.11.0]` — they lead **Changed** and **Fixed**.
- Dropped the two generated one-liners they supersede (`#542`, `#543`) — exactly two,
  asserted in the edit script.
- Added a **Documentation** entry for #575 (the 12 previously-undocumented published
  packages).
- Emptied `[Unreleased]`.

`releaseNotes` extracts this section verbatim into the GitHub Release body, so the
Release will match the changelog rather than a commit list.

## Verification

- `npm test` green **before** the bump — 54/54 segments
- `npm test` green **after** bump + changelog — 54/54 segments (R6: this is the tree
  that gets tagged)
- **R4** — every one of the 47 manifests, `package-lock.json`, and both marketplace
  manifests read `1.11.0`; no stale `1.10.0` remains. (v1.1.0 stranded the lockfile;
  v1.1.0 is why this is checked.)
- **R7** — `packages/core/package.json` (the `versionSource`) reads exactly `1.11.0`
- Diff outside `CHANGELOG.md` is **version strings and `@adlc/*` repins only** — verified
  by filtering the diff
- Conformance checker: **conformant** (gated `publish` job bound to `npm-publish` with a
  required reviewer, `id-token: write`, no repo/org token; the environment-scoped
  bootstrap token passes under the profile's declared `tokenException`)

## After merge

The tag is pushed **only** after re-validating the merge commit, and the publish then
waits on the `npm-publish` reviewer gate. I will not approve that deployment.


voodootikigod:
## Green — 19/19, including `gate`

### The changelog defect the review caught

The first distinct-provider review flagged a **real** defect I introduced:
`## [Unreleased]` ended up *below* `## [1.11.0]` instead of at the top of the file.

Root cause is `scripts/changelog.mjs` — it inserts the new version section *above*
the existing `[Unreleased]` rather than folding it in, and my curation preserved
that order. It matters beyond tidiness: the next release's `[Unreleased]` entries
would accumulate *between* 1.11.0 and 1.10.0, breaking reverse-chronological order
and the section-slicing that `changelogCommand` and `releaseNotes` both rely on.

Fixed — `[Unreleased]` is back at the top, `[1.11.0]` directly below it. Worth a
follow-up ticket against `changelog.mjs` so the next release does not repeat it.

### Cross-model attestation

A lockstep bump necessarily touches every `packages/**` and `plugins/**` manifest,
so it is TRUST-ROOT tier via nine completed tickets' rails (T37, T38, T18, T42,
T43, T44, T49, T52, T53). Every release bump will be.

- **Review:** `agy` (distinct from author `anthropic`), `--verify`, on the final tree
- **Verdict:** APPROVE — 48 files examined, `mode: inline-diff`, no material findings
- **Revision:** `git-change:fce51a65:8a38221d` — recomputed after the changelog fix
  and confirmed to match what the review actually covered
- **Signed:** `sigVersion: 2`, `signed: true`
- **Ticket:** `T-01M0TC5BP3PRR08TP38FP8PV6C` — created for this release rather than
  reusing an unrelated one, as the 1.10.0 bump did

Verdict recorded is the verdict received. No `approve` was recorded over a
`needs-attention`; the one finding was fixed and the review re-run clean.

### Verification

- `npm test` green at every stage — pre-bump, post-bump/changelog, post-fix,
  and on the exact tagged tree: 54/54 segments each time
- **R4** — 47 manifests + lockfile + both marketplace files at `1.11.0`; no stale `1.10.0`
- **R7** — `versionSource` reads exactly `1.11.0`
- Diff outside `CHANGELOG.md` is version strings and `@adlc/*` repins only

### Next (R1 — not mine to do)

**Merge is yours.** After merge I re-validate the merge commit, tag `v1.11.0`, and
hand you the run URL. The publish then waits on the `npm-publish` reviewer gate —
**I will not approve that deployment.**

Issue #162, opened 2026-07-12
CC rail hook unions ALL tickets' rails (incl. completed) — blocks legitimate new work; inconsistent with rails-guard CLI & opencode/pi


voodootikigod:
## Summary

The Claude Code in-session rail hook (`plugins/adlc-claude-code/hooks/adlc-hook.mjs`, `rails()`) **unions the `rails` of every ticket in `.adlc/tickets.json`** — including tickets already marked `completed: true` — and enforces that union unconditionally, with **no active-ticket narrowing** and **no `ADLC_P4_ENFORCEMENT` gate**. As a result, a *completed* ticket that declared a broad rail permanently freezes those paths in every future session, blocking legitimate new work the current ticket explicitly authorizes.

This behavior is **inconsistent with the other two enforcement layers** (the `rails-guard` CLI and the opencode/pi in-session checkers), both of which narrow to a single ticket. It also appears to contradict the `completed:true` lifecycle invariant (#104), which states that backlog consumers must filter completed tickets.

## Concrete reproduction (observed while building `@adlc/fleet`, T42)

1. Ticket **T37** (`completed: true`) declared `rails: ["packages/**", "docs/ci/**", ...]`. Ticket **T38** (`completed: true`) declared `rails: ["docs/adr/**", "packages/**"]`.
2. Ticket **T42** is authored with `scope: ["packages/fleet/**", "docs/adr/0010-...md", ...]` and `rails: ["packages/core/**"]` — i.e. T42 explicitly authorizes creating `packages/fleet/**`, and its *own* frozen rail is only `packages/core/**`.
3. A `Write` to the brand-new `packages/fleet/package.json` is **denied** by the hook:

   > `packages/fleet/package.json is a frozen rail declared by ticket T37 (rails: "packages/**"). Edits to frozen rails are blocked during build.`

4. The only way forward is `ADLC_RAILS_BYPASS=1` — an audited bypass — even though **no genuinely protected invariant is touched** (`packages/core/**`, T42's real rail, is untouched; verified independently by `rails-guard --ticket T42`, which **passes** because the CLI narrows to T42).

So the CLI gate and the in-session hook **disagree on the same change**: `rails-guard --base main --ticket T42` → exit 0 (clean), while the in-session hook → DENY.

## Root cause — three layers, three different rail-resolution rules

| Layer | File | Rail set it enforces | Enforcement gate | Completed filter |
|---|---|---|---|---|
[...]

Issue #59, opened 2026-07-02
feat(plugins): mechanical adversarial-review trigger on the risk gate (Stop-hook verdict check)


voodootikigod:
## Context

`adversarial-review` (the multi-provider adversarial code review CLI, `voodootikigod/adversarial-review`) is never triggered mechanically anywhere in this repo — no hook, no CI job, no pre-commit. Confirmed by grepping `.github/`, `plugins/*/hooks*`, and `scripts/` for the string `"adversarial-review"`: zero matches outside documentation.

- `plugins/adlc-claude-code/hooks/adlc-hook.mjs` (759 lines) implements exactly four modes: `preflight` (SessionStart → `adlc preflight`), `flail` (PostToolUse → `adlc flail-detector`), `manifest` (Stop → `adlc gate-manifest verify`), `rails` (PreToolUse → path-glob enforcement). None reference `adversarial-review`.
- `plugins/adlc-opencode/lib/session-hooks.mjs` (69 lines) implements `checkPreflight` and `auditGateManifest` — same story, no reference.

This was a deliberate, documented deferral, not an oversight:

- `docs/adr/0005-adversarial-design-review-gate.md:5-8`: *"Adopt the gate as a recommended, risk-gated P1 practice ... do **not** yet build mechanical P2-entry enforcement into the `coldstart`/router path. Revisit the tooling investment once trust-boundary features are built often enough that relying on the operator to invoke it proves insufficient."*
- `docs/adr/0007-multimodel-adversarial-review.md:3`: *"Status: Accepted — recommended practice (mechanical enforcement deferred)."* Line 83-84: *"mechanical enforcement in the gate path is **deferred** until the manual practice proves insufficient."*

A multi-agent review of this toolkit (2026-07-02) is exactly that signal: adversarial-review is real, useful, and consistently skipped because nothing forces it.

## Evidence

- Risk-tier path patterns are already defined and ready to key off: `docs/adr/0007-multimodel-adversarial-review.md:42-47` — authentication/authorization/trust boundary; a security control or deny path (rail guards, validators, sandboxes); secrets handling; data-loss/destructive/irreversible operations; schema or migration changes; CI/CD or supply-chain config.
[...]

PR #577, opened 2026-08-24
feat(tickets): record a signed audit entry for every bypassed store mutation; refuse keyless bypass writes

voodootikigod:
Builds ticket **T-01M0122WMF8EJTB7ERHTEG8HMJ** — "Record a signed manifest audit entry for every bypassed ticket-store mutation".

## The gap

Once any ticket declares a rail, the ticket store is a **frozen trust root**: it is the configuration that decides what the rail guards freeze, so freezing it too is the only thing that stops one edit from disabling enforcement. The PreToolUse rail hook has always denied a structured edit to `.adlc/tickets/**` from that moment on unless the override is recorded to the gate-manifest, and `/adlc:adlc-ticket` told operators that "editing the ticket set while rails are frozen is a deliberate, audited action."

The CLI reaches the same bytes through a door the hook deliberately does not parse — a Bash command — and arrived unaudited.

### Empirical before/after

Throwaway repo, one rails-declaring ticket in the store, `ADLC_RAILS_BYPASS=1`, no `ADLC_MANIFEST_KEY`:

| | `adlc ticket create --write` |
|---|---|
| **Before** | exit 0 · store **MUTATED** · **zero** manifest entries |
| **After** | refuses (`MANIFEST_KEY_REQUIRED`) · store byte-identical · no pending transaction · nothing recorded |

With a key the same command applies the mutation and appends exactly one signed `ticket-mutation` entry whose `storeHashAfter` matches `ticket store status --json`. Both halves are re-run as the first test in the new suite.

## What changed

Enforcement lives in the library every door goes through, not in a bin a different caller can route around.

- **One entry per mutation.** Mutating a frozen trust root appends exactly one manifest entry carrying `{ op, ticketId(s), storeHashBefore, storeHashAfter, bypass: true }`. A mutation with no gate of its own records under `ticket-mutation`; a sensitive one (rail narrowing, completion, reassignment, archive, migration) keeps its existing `ticket-<operation>` entry and gains the audit fields.
- **Fail closed on the key**, following the #370 `--allow-unsigned` precedent: with no key the write refuses *before* anything is touched — no store change, no staged file, no pending transaction, no archive directory. `--allow-unsigned` records an unsigned entry deliberately and warns.
- **Every door.** The contract now covers `adlc ticket create/update/complete/discard/edit`, archive/restore, the legacy→sharded migration **and its recovery**, recovery of an interrupted transaction, `ticket-prune`'s direct `tickets.json` write, and `ticket-sync` pull/push. Each of those was found still open by cross-model review and is closed with a test.
- **The trust root is sticky.** The predicate reads the active store, the archive, **and the append-only manifest**. Archiving, discarding, or un-railing the last railed ticket therefore cannot switch the audit off for everything after it (#162) — the mutation that removes the last rail records its own override on the way out, and the manifest keeps it.
- **`export` is no longer a writer into `.adlc/`.** It refused nothing before; it now refuses any target resolving into `.adlc/` (the evidence ledger included) or onto the source store, through symlinks at any depth.
- **`LegacyTicketStore.write` is now audited.** It was a declared public method that replaced the whole store with no rails check, key, or evidence. It is now a compatibility adapter over `applyLegacyTransaction`: the published 1.x one-argument call keeps working unchanged on a rails-free store, and on a trust root it refuses without a key instead of writing unaudited.
- **Zero ceremony where none is owed.** A repo with no rails anywhere needs no key, records nothing, and every pre-existing caller's evidence bytes are unchanged.

## Design notes

- **The trigger is store state, not the env var.** The ticket's build note says "performed with `ADLC_RAILS_BYPASS` set (i.e. the store is a frozen trust root and the write is a bypass)" and its ACs say "with rails frozen". Keying off the environment variable would have made the audit trivially evadable by the mutating party. Mutating a frozen trust root *is* the bypass.
- **The journal is not evidence.** Recovery decides "is an audit owed" from the repo, from the journal's hash-verified *backups* (which still show a rail even when the transaction being recovered removed the last one), and from the journal's flag — OR-ed, so the flag can only add. Editing `bypassAudit: true` → `false` changes nothing.
- **`stage → audit → commit` in `ticket-prune`.** That path has no journal, so the store write and the manifest append cannot be atomic. Recording after a completed write can leave a real mutation with no record — undetectable afterwards. So the content is staged, the entry appended, and only a same-directory rename remains; a failed audit discards the staged copy and refuses with the store untouched. Each mutation gets a **fresh** transaction id rather than one derived from the before/after hashes: a derived id cannot tell a retry from a genuine repeat, so reverting a tombstone and re-applying it would compute the same id and be waved through as an old retry — a real change to a trust root with no record. A failed rename instead appends a **compensating** entry (`action: 'abandoned'`, before and after hash equal) saying the store did not move; if that append fails too, the refusal says the ledger stands uncorrected rather than letting the caller assume it is true.
- **Idempotency covers the audit payload.** A retry may skip writing evidence only if the existing entry matches on it too; otherwise an entry lacking `bypass`/`storeHashBefore` would wave a mutation through as "already audited". One narrow exception, used by the migration replay alone: an entry carrying *none* of the fields is a pre-feature record that can never grow them, and demanding them would wedge that recovery permanently.
- **`ticket-sync push` preflights before its remote calls.** Not the enforcement point — the transaction re-checks under the lock — but it removes the common case where a push creates issues upstream and then refuses locally.

## Operational consequence (deliberate)

In a repo that declares rails, these now require `ADLC_MANIFEST_KEY` or an explicit `--allow-unsigned`: `adlc ticket <mutation> --write`, `adlc ticket store migrate/recover`, fleet completion, `ticket-sync pull/push`, and `ticket-prune --write`. The ceremony-drift issue body now tells the admin to export the key, since it would otherwise print a command that cannot work.

Verified that the **real** `rails-guard-ci` still exits 0 on a synced and on a pruned railed store with the audit entry present, given the bootstrap-created empty manifest at base.

## Test plan

- `packages/tickets/test/bypass-audit.test.mjs` (new, 56 tests) — the empirical probe first, then AC1–AC4, the key-contract fail-closed cases, every other door into the store, the tampered-journal and stripped-journal cases, the archive/discard/un-rail thaw cases, the export guard (including nested symlinks), and idempotency.
- Fail-closed defaults are pinned wherever only a library caller can reach them — `TicketService`, `recoverMigration`, `pull`, and the `ticket-prune` bin each have a test that fails if the default flips.
- Downstream suites updated where a fixture's store is a trust root (fleet, ticket-sync, ticket-prune, prosecute, pi, ceremony-drift), each with a comment saying why the key is incidental to what that test asserts. `ticket-sync` gains the other half of its invariant: a sync against an unrailed store still emits nothing.
- `scripts/test/roundtrip-coverage.test.mjs` learns the staged-write spelling, so splitting an atomic write in two cannot take a package out of the writer-boundary gate's sight.
- `npm test` — 54/54 segments green, run with `ADLC_MANIFEST_KEY` unset so it matches CI rather than a developer shell that exports one.
- `npm run preflight` — green, mutation-gate 0 survivors.

Rebased onto `origin/main` after #571 merged. Completion follows separately.

## Cross-model review: 27 rounds, one standing finding

Codex (`--provider codex`), foreground, re-run on the exact tree after every change. Roughly 35 findings were fixed, including several that were real holes in the first implementation: the migration, archive/restore, transaction recovery, `ticket-prune` and `ticket-sync` doors were all still open; the recovery predicate trusted an editable journal; `export` could overwrite the evidence ledger; `LegacyTicketStore.write` was an undocumented unaudited writer; a converged sync would have demanded a key and appended a phantom entry.

**Not attested — deliberately.** No cross-model `approve` is recorded for this revision, so `gate` is red. Minting an attestation I do not have would be exactly the class of unearned evidence this ticket exists to prevent. **Every other check is green.**

### The standing finding: an ACCEPTED, tracked residual

> A repo whose last railed ticket was removed by the PRE-change writer has no rail in the ticket set, nothing in the archive, and no manifest marker, so this code reads it as a repo that never used rails.

Adjudicated, and **accepted by the maintainer** as a documented, bounded, historical-only residual. It is deliberately **not** closed here; its closure is tracked by ticket **T-01M0TMRSQKGTNWZTFXBPQ2JHNB** (PR #578 — the adoption-record / manifest-key-hermeticity mechanism), so this is a decided and owned limitation rather than a gap someone forgot.

Why accepting it is safe: the evidence does not exist to read — that writer recorded nothing — so closing it needs either a durable adoption record or git archaeology over `.adlc/tickets` on every mutation, which does not work outside a git checkout. The residual is historical-only and **cannot grow**: every removal from this change forward leaves a marker, so the uncovered set is fixed at whatever pre-change repos already exist and shrinks to nothing as each takes a single audited write. It is a historical-state limitation, not a bypass an attacker can arrange in a repo that upgrades with rails present.

The same framing is now recorded in the code at `packages/tickets/lib/trust-root.mjs` ("KNOWN BOUNDARY"), in `packages/tickets/README.md`, and on the toolkit docs page.

One further recurring finding is a deliberate, documented position rather than an oversight:

- **`ticket-sync push` preflights before the store lock.** It is an early-out that removes the common partial-failure, not the enforcement point; the transaction re-checks under the lock. Holding the store lock across every network round trip would block all other writers for the length of a remote sync.


voodootikigod:
P5 evidence (final, run by team-lead on the exact head c0e055df since the lane's round didn't deliver): codex, full diff inlined (--max-bytes 600000), 9 files examined. **findings[] = 1**, and it is an ALREADY-ACCEPTED, ticket-tracked residual:

- prune crash-window false-audit (medium 0.99) → **T-01M0WNX6P09HWKDK429XQ8GGRJ (#580)**, accepted by the maintainer as a bounded, detectable, pre-existing residual (real fix = durable journal, a feature).

The [2] regression (LegacyTicketStore.write inferring root → keyless unaudited external write) is GONE. Finding [3] (export TOCTOU → T-01M0WNX6ZA0D94HW2VQKZAPGQ2, #580) did not resurface this round; tracked regardless. Historical KNOWN BOUNDARY → T-01M0TMRSQKGTNWZTFXBPQ2JHNB (#578, MERGED).

Note: the JSON `verdict` field reads 'approve' while the summary says do-not-ship over that one medium — the known verdict/exit-code inconsistency (issue #59); only findings[] is authoritative, and here that one finding is an accepted residual.

CI: green except `gate` (the accepted boundary red; admin-merge is the sanctioned path). So every standing item on this PR maps to a tracked ticket. **Merge order: #580 first (this PR's code cites its two ids), then admin-merge #577 over the gate red.**

PR #581, opened 2026-08-25
chore(adlc): complete the four Wave-2 open-ticket fixes

voodootikigod:
Completion batch for the post-release Wave-2 tickets, all shipped and prosecuted:

| Ticket | Fix | PR |
|---|---|---|
| T-01M0122Z30… | wrapper import-time fail-open deny | #565 |
| T-01M0DF85FK… | quartermaster docs threat-model truth | #566 |
| T-01M0122YKB… | add-vs-alter merge-base anchor | #571 |
| T-01M0122WMF… | bypassed-mutation signed audit | #577 |

All railless T36 completion annotations (authorized, audit segment included); gate-clean. Their three accepted residuals are tracked in #578/#580.


voodootikigod:
### CI fix — `test` matrix (18/20/22) was red on one test

**Failing test:** `packages/tickets/test/bypass-audit.test.mjs` — *"the exported trust-root guard fails CLOSED on an omitted or invalid key"*. Its last assertion expected a rails-free fixture store **not** to be a trust root, but called `assertSignableTrustRootWrite` with the default `root: '.'` — the process cwd, which under `npm test` is this repository.

**Why it only broke here:** `repoDeclaresRails` also consults the repository's own `.adlc/manifest.d` (a recorded bypass is a sticky trust-root witness). This PR is the first change to land `"bypass": true` audit entries there — the four audited `ticket complete` writes that #577 built. From the repo root that witness now answers *true*, so the guard refused the fixture with `MANIFEST_KEY_REQUIRED`. `main` passed only because its manifest held no bypass entry yet; a `.adlc/`-only PR with zero code changes was enough to expose it.

**Fix (09c0126e):** every call in that test that can reach the repository now names an empty temp root, so it exercises the key contract against the fixture and nothing else. Verified locally: the test and the full `packages/tickets` suite pass (419/419), the guard is still load-bearing (an early return in `assertWriteIsSignable` fails this test and the AC tests around it), and `npm run preflight -- --base main` passes all 6 gates. `bypass-audit.test.mjs` is under no active rail.

PR #879, opened 2026-08-26
fix(gates): harden Wave 2 gates against false greens (#768, #775, #786, #823)

### LINKED ISSUE #768: skill-rot: an explicit path argument that does not exist is silently ignored instead of an operational error
**Source:** `/release-audit` for @adlc 1.11.0 (`release-blockers.md` @ `60050b856`), artifact `pkg:skill-rot`, bucket **SHOULD-FIX**, class **false-green**.
**Re-verified** against `main` @ `5b3914a2` on 2026-08-25: still present. Unchanged since audit. Explicit positionals and default roots share the skip-if-absent path; a missing explicit path is only an error when no root yields any skill.

**Location** `packages/skill-rot/lib/find-skills.mjs:22`
**Blocker test at audit** ✓ user hits it · ✓ needs another release · ✕ worse than status quo

```
    if (!existsSync(absRoot)) continue;
```

findSkills skips any root that does not exist. That is correct and documented for the three DEFAULT roots ("searched only when they exist"), but positionals go through the same code path with no distinction. A user who passes an explicit path — a typo, a directory that was renamed, a CI path that moved — gets no error at all.

Empirically confirmed:

  $ skill-rot good typo-dir-that-does-not-exist
    [OK]    good/a/SKILL.md
  Summary: 1 skill(s) checked, 1 clean, 0 stale
  EXIT=0

The operator explicitly named a subject; the gate did not check it and did not say so. It only surfaces as exit 1 in the degenerate case where NO root resolves to any skill.

**Consequence.** A CI job pinned to a skills directory that was later moved keeps passing green forever, having checked nothing in that tree.

**Recommendation.** Thread a flag distinguishing default roots from explicit positionals. For explicit positionals, a missing path is an operational error (exit 1) with the resolved path named; for defaults, keep the current skip-if-absent behaviour.

---
Filed from the 1.11.0 release audit's non-blocking findings. Current evidence: `20-22: for (const root of roots) { const absRoot = resolve(repoRoot, root); if (!existsSync(absRoot)) continue;  |  bin/skill-rot.mjs:29 const searchRoots = positionals.length > 0 ? positionals : DEFAULT_ROOTS;`

https://claude.ai/code/session_015HEKHKtnqc1Q1R1HesdSPH

### LINKED ISSUE #775: spec-lint: --record comma-splits the spec path via rawFiles, writing an entry with null file hashes and exiting 0
**Source:** `/release-audit` for @adlc 1.11.0 (`release-blockers.md` @ `60050b856`), artifact `pkg:spec-lint`, bucket **SHOULD-FIX**, class **data loss**.
**Re-verified** against `main` @ `5b3914a2` on 2026-08-25: still present. Unchanged since audit (no commits to spec-lint/lib or gate-manifest/lib). A single known spec path is still routed through the comma-splitting --files parser, producing null file hashes and exit 0.

**Location** `packages/spec-lint/lib/record.mjs:29`
**Blocker test at audit** ✕ user hits it · ✕ needs another release · ✓ worse than status quo

```
  return record({ gate: GATE_NAME, ticket, rawData, rawFiles: specPath, dir, key });
```

`recordResult` hands the spec path to gate-manifest's `record()` as `rawFiles`, which is the raw `--files` CLI string — `parseFileList` splits it on commas. A spec path containing a comma is torn into fragments, `hashFiles` swallows the resulting ENOENTs (`catch { out[p] = null; }` in core/lib/ledger.mjs) and the run still exits 0. Reproduced:

  $ spec-lint 'spec,v2.md' --record --ticket T1 --dir .adlc  ->  exit=0
  {"seq":1,"gate":"spec-lint",...,"data":{"verified":true},"files":{"spec":null,"v2.md":null},"prev":null}

The entry claims `verified: true` for a ticket while binding to no real file content — the exact P1 D4 binding the record was added to provide. It is not a false green end-to-end: the runner's P1 assertion rejects it (`if (auditPaths.length !== 1) return false` in packages/runner/lib/assertions.mjs), so the operator lands on "P1 evidence is contradictory" much later with no hint that a comma caused it. New in this release (`lib/record.mjs` did not exist at v1.10.0).

**Consequence.** A comma anywhere in the spec path produces a silently useless manifest entry; the operator only discovers it at the P1 gate as an unexplained "does not audit the approved spec file" error.

**Recommendation.** Do not route a single known path through the comma-splitting raw-string parser. Either call the structured append with `filePaths: [specPath]`, or reject a spec path containing a comma with an `opError` before recording.

---
Filed from the 1.11.0 release audit's non-blocking findings. Current evidence: `record.mjs:29 return record({ gate: GATE_NAME, ticket, rawData, rawFiles: specPath, dir, key });  |  packages/gate-manifest/lib/record.mjs:230 const filePaths = parseFileList(rawFiles);  |  :159-162 export function parseFileList(raw) { if (!raw) return []; return raw.split(',').map(s => s.trim()).fi`

https://claude.ai/code/session_015HEKHKtnqc1Q1R1HesdSPH

### LINKED ISSUE #786: ticket-sync: bin is a silent no-op through the npm .bin symlink (main-module guard never matches) — doctor exits 0 having run nothing
**Source:** `/release-audit` for @adlc 1.11.0 (`release-blockers.md` @ `60050b856`), artifact `pkg:ticket-sync`, bucket **SHOULD-FIX**, class **false-green**.
**Re-verified** against `main` @ `5b3914a2` on 2026-08-25: still present. #577 touched this bin but not the main-module guard; the symlink realpath mismatch still makes every command through .bin / npx / global install a silent exit-0 no-op. packages/fleet/bin/fleet.mjs:185 uses the weaker `file://${process.argv[1]}` form and has the same defect.

**Location** `packages/ticket-sync/bin/ticket-sync.mjs:144` (now `packages/ticket-sync/bin/ticket-sync.mjs:155`)
**Blocker test at audit** ✓ user hits it · ✓ needs another release · ✕ worse than status quo

```
if (process.argv[1] && import.meta.url === pathToFileURL(process.argv[1]).href) {
```

The main-module guard compares `import.meta.url` (which Node resolves to the file's REALPATH) against `pathToFileURL(process.argv[1])` (which Node leaves as the path actually invoked). npm installs `adlc-ticket-sync` as a SYMLINK at `node_modules/.bin/adlc-ticket-sync -> ../@adlc/ticket-sync/bin/ticket-sync.mjs`, so the two never match and `main()` is never called.

I confirmed this against the REAL PUBLISHED TARBALL, not the working copy. In a clean scratch dir: `npm i @adlc/ticket-sync@1.10.0` then

  ./node_modules/.bin/adlc-ticket-sync --help   -> no output, exit 0
  ./node_modules/.bin/adlc-ticket-sync doctor   -> no output, exit 0
  npx adlc-ticket-sync --help                   -> no output, exit 0
  node ./node_modules/@adlc/ticket-sync/bin/ticket-sync.mjs --help  -> works

The sibling `./node_modules/.bin/adlc-tickets --help` from the same install prints its usage normally, so this is not an environment quirk — it is specific to this bin's guard. Isolated repro on a non-symlinked path (Node v22.21.1): invoking a script through a symlink gives `argv1=/…/sub/link.mjs`, `meta=file:///…/real.mjs`, `match=false`.

This is the signature failure this product class must not have: `doctor`, whose contract is `0 ok · 1 operational · 2 blocked`, returns 0 having executed zero checks. Any CI step wired as `npx adlc-ticket-sync doctor` is a permanently green gate that prosecutes nothing.

Scope of breakage: BROKEN for `npm i -g` + `adlc-ticket-sync …` (the README's own quickstart), local `npm i` + `.bin`, and `npx`. WORKS for `adlc ticket <pull|push|sync|doctor>` via @adlc/cli, because `packages/cli/lib/dispatch.mjs` resolves the package.json realpath and joins the bin path (no symlink in the invoked argv).

Not a blocker only because v1.10.0 shipped the identical code — this artifact is unchanged. It is nonetheless the highest-value one-line fix available in this release.

**Consequence.** A user who follows the README (`npm i -g @adlc/ticket-sync; adlc-ticket-sync --help`) or wires `npx adlc-ticket-sync doctor` into CI gets total silence and exit 0 from every command — a gate that can never fail, and a sync that never syncs.

**Recommendation.** Resolve the symlink before comparing: `import { realpathSync } from 'node:fs'` and use `pathToFileURL(realpathSync(process.argv[1])).href`, guarded in a try/catch. Add a test that invokes the bin through a symlink and asserts non-empty stdout, so the idiom cannot silently rot again. Note packages/fleet/bin/fleet.mjs uses an even weaker variant (`file://${process.argv[1]}`) and likely has the same defect — worth a suite-level sweep.

---
Filed from the 1.11.0 release audit's non-blocking findings. Current evidence: `if (process.argv[1] && import.meta.url === pathToFileURL(process.argv[1]).href) {  — live repro in this checkout: ./node_modules/.bin/adlc-ticket-sync --help → 0 bytes stdout, exit 0; …/.bin/adlc-ticket-sync doctor → 0 bytes, exit 0; node packages/ticket-sync/bin/ticket-sync.mjs doctor → 422 bytes, `

https://claude.ai/code/session_015HEKHKtnqc1Q1R1HesdSPH

### LINKED ISSUE #823: plugin(adlc-gemini): rails guard allows on empty stdin and non-object JSON under enforcement
**Source:** `/release-audit` for @adlc 1.11.0 (`release-blockers.md` @ `60050b856`), artifact `plugin:adlc-gemini`, bucket **SHOULD-FIX**, class **false-green**.
**Re-verified** against `main` @ `5b3914a2` on 2026-08-25: still present. Reproduced at HEAD: empty/blank stdin and JSON scalars still fall through to the 'other' no-path allow branch while malformed JSON denies.

**Location** `plugins/adlc-gemini/hooks/adlc-rails-guard.mjs:171`
**Blocker test at audit** ✕ user hits it · ✓ needs another release · ✕ worse than status quo

```
  if (raw && raw.trim()) {
```

`runFromStdin` only enters the parse path when `raw` is truthy and non-blank. Empty stdin therefore leaves `payload = {}`, which yields tool name `''` -> `classifyTool('') === 'other'` -> no paths -> the `'other'` no-path allow branch. A payload of `null` or `"hi"` parses fine and lands in the same place.

Probed against the shipped shim with ADLC_P4_ENFORCEMENT=1:
- `printf '' | node hooks/adlc-rails-guard.cjs` -> `{"allow_tool":true}`
- `printf '   ' | ...` -> `{"allow_tool":true}`
- `printf 'null' | ...` -> `{"allow_tool":true}`
- `printf '"hi"' | ...` -> `{"allow_tool":true}`
- `printf '{nope' | ...` -> `{"allow_tool":false, ... unparseable tool payload while enforcing}`

So the guard fails CLOSED on garbage but fails OPEN on nothing at all, and on any well-formed non-object. If the host ever delivers an empty or non-object payload (a stdin race, an EPIPE, a payload-shape change on the agy side), enforcement is silently off for that call and the hook still reports success. `test/shim.test.mjs` pins the malformed-JSON pair but nothing pins the empty/absent case.

**Consequence.** Under active enforcement, a tool call whose payload arrives empty or as a JSON scalar is waved through with no deny and no diagnostic, so an operator sees a healthy hook while the gate did nothing.

**Recommendation.** Treat absent/blank stdin and a parsed non-object payload the same as unparseable input: deny under enforcement, allow when off. Add tests for `''`, `'   '`, `'null'`, and `'"hi"'` alongside the existing malformed-JSON pair in test/shim.test.mjs.

---
Filed from the 1.11.0 release audit's non-blocking findings. Current evidence: `if (raw && raw.trim()) { — repro with ADLC_P4_ENFORCEMENT=1: printf '' | node hooks/adlc-rails-guard.cjs -> {"allow_tool":true}; 'null' -> allow; '"hi"' -> allow; '{nope' -> {"allow_tool":false,...unparseable tool payload while enforcing}`

https://claude.ai/code/session_015HEKHKtnqc1Q1R1HesdSPH

voodootikigod:
## Summary

This PR addresses 4 P1/high false-green and operational bugs identified during gate reliability audits:

### 1. Issue #768 — : Explicit non-existent search roots fail with operational error
- **Problem**: When explicit positional search roots were provided on the CLI (e.g. `skill-rot non-existent-dir`), `findSkills` silently skipped missing roots instead of failing. If other default roots had skills, it exited 0.
- **Fix**: Added `strict` option to `findSkills` and explicit positional validation in `bin/skill-rot.mjs` to output an operational error and exit with code 1 when an explicit path does not exist. Added support for checking single `SKILL.md` file paths.

### 2. Issue #775 — : Spec paths containing commas hashed without fragmentation
- **Problem**: `spec-lint --record` passed `rawFiles: specPath` to `gate-manifest`'s `record()`, which comma-split the string via `parseFileList`. A spec file containing a comma (e.g., `spec,v2.md`) was fragmented into non-existent path keys with `null` file hashes.
- **Fix**: Updated `packages/spec-lint/lib/record.mjs` to invoke `appendManifestEntry` directly with `files: specPath ? hashFiles([specPath]) : {}`, avoiding string-splitting.

### 3. Issue #786 —  & : Resolve symlinks in `isMain()`
- **Problem**: Direct comparison of `process.argv[1]` with `import.meta.url` failed when binaries were invoked via symlinks (such as npm `.bin` links), causing commands like `adlc ticket doctor` or `fleet` to exit 0 without executing anything.
- **Fix**: Wrapped the main check in `isMain()` using `realpathSync(process.argv[1])` inside a try/catch block.

### 4. Issue #823 — : Rails guard fails closed on empty stdin and non-object JSON payloads
- **Problem**: When `ADLC_P4_ENFORCEMENT=1`, empty/whitespace stdin or JSON scalar values (`null`, `"hi"`, `123`, `[]`) fell through to an unclassified tool name and were permitted.
- **Fix**: Updated `plugins/adlc-gemini/hooks/adlc-rails-guard.mjs` to fail closed on empty/blank stdin and non-object/array JSON payloads under enforcement, returning `{ allow_tool: false, deny_reason: ... }`.

---

## Adversarial Review Verdict
- **Verdict**: `approve`
- **Lenses Evaluated**: Correctness/Red-team, Simplicity/YAGNI, Security/Privacy, Evidence Runner, Anti-slop (all PASS).
- **Test Suite**: 400/400 tests passing across all modified packages and plugin shims.

Closes #768, Closes #775, Closes #786, Closes #823.


voodootikigod:
## Cross-model prosecution + attestation (Claude Fable 5 author, codex reviewer)

**Verdict:** ship — round 3 of `npx adversarial-review --base origin/main --provider codex` returned a genuine `approve` (0 findings, model verdict = approve, no grounding warnings). Signed attestation recorded (`openai` vs author `anthropic`) @ `git-change:7e319ac6…:ea6637c2…`; local `tier-check --author-provider anthropic` → **PASS**.

### CI reds addressed
- **mutation-gate**: CI's 4 survivors (fleet/ticket-sync `isMain()` bool-flip + off-by-one, gemini `decide()` logic-swap) were the visible part — a deeper local run (`--max 40`) found 6 more. Now **36/36 killed**. Notable: two survivors were *equivalent* mutants (a `return false` in a boolean-context catch; a defaulted `{ strict = false }` every caller overrides) and were removed by restructuring rather than by bolting on hollow tests, per the gate's own guidance.
- **cross-model gate**: attested as above.

### Findings fixed in-PR (the body's "approve, all lenses PASS" predates these)
| Source | Finding → fix |
|---|---|
| mutation gate | fleet's #786 fix shipped with **no test**: added symlinked-argv[1], no-argument dispatch, `node -e` bare import (no / nonexistent argv[1]) regressions for fleet and ticket-sync |
| mutation gate | `isMain()` → `entryUrl()` (realpath'd `file://` href or `null`) compared at the call site |
| mutation gate | gemini: one shared `isToolPayload()` predicate for `decide()` and `runFromStdin()`; direct `decide()` scalar/array/null tests |
| mutation gate | spec-lint `recordResult`: parity test vs gate-manifest `record()` under a key (same fields, v1 sig, hashes) — the invariant of the #775 refactor |
| mutation gate | skill-rot strict: an explicit path that is neither a directory nor `SKILL.md` is an operational error, not a silent skip |
| codex r1 | gemini: `{}` / `{toolCall:{}}` (no tool name) classified "other" and were **allowed** under enforcement — the same fall-through #823 closes for scalars, one step later → fail closed |
| codex r1 | skill-rot: `collectSkills` swallowed `readdir` failures under explicit roots → strict threads into the walk; unreadable subtree = exit 1 |
| codex r2 | gemini: a nameless envelope was recorded against the session (depth) **before** `decide()` denied it — repeated malformed envelopes could push a victim session over the depth threshold → short-circuit to `decide()` before any tracker write; regression pins depth 0 with a named positive control at 1 |

Follow-up filed: #899 — `adlc-opencode/bin/cli.mjs` has the same un-realpath'd entry check (kept out of this PR to hold its declared scope).

Verification: touched suites green every round; final `node scripts/run-tests.mjs packages` → 31/31 segments.

Issue #1003, opened 2026-09-12
tickets: no atomic multi-ticket write path — looping create is order-fragile and non-atomic


voodootikigod:
## Summary

There is no way to write a multi-ticket DAG into the store atomically. `adlc ticket create --input` takes exactly one ticket, and the only workaround is looping it in reverse-topological order — non-atomic, and it leaves a partial DAG behind if it fails midway.

This came out of specification work for a P2 decomposition compiler (`.adlc/specs/decompose.md`), which needs to land 4–8 related tickets at once. The compiler itself was descoped to emit JSON rather than write, precisely because this contract turned out to be the hard part. Recording the constraints here so they don't have to be rediscovered.

## Why looping `create` is not sufficient

Edges live on the **blocking** ticket and name its dependent (`{to: <id>}`), and `packages/tickets/lib/schema.mjs:49` rejects `edge to unknown ticket`. So a multi-level DAG can only be created leaf-first, in reverse topological order. That works, but:

- it is order-fragile and not something a human gets right by hand;
- there is no transaction, so a failure at ticket *k* leaves *k-1* tickets in the active set;
- each create appends its own signed audit entry, so one logical operation produces N unrelated-looking entries.

## Constraints any implementation must satisfy

These are the findings from three adversarial premortem rounds against the compiler spec. Each is grounded in current code.

**1. `planReconciliation` is the wrong tool.** It hard-codes `operation: 'remote-reconciliation'` (`packages/tickets/lib/service.mjs:202`), so every batch write would be audited as a ticket-sync event. Its full-snapshot-replace shape also risks dropping tickets if the snapshot goes stale mid-run.

**2. A new `planCreateBatch` must replicate `planCreate`'s guards.** `planCreate` calls `#assertIdNotArchived` (`ARCHIVE_COLLISION`) and `#assertNoManifestRails` (`service.mjs:96`, `RAIL_COVERS_MANIFEST`, #235) before mutating. A batch path that skips them opens an unguarded write door past #235.

**3. Per-ticket evidence binding is NOT reachable without a redesign.** This is the subtlest one:

- `recordTicketEvidence` derives `bindingScope: ticketId ? 'ticket' : 'store'` from a **singular** id (`packages/tickets/lib/evidence.mjs:238`), so a batch write records one entry naming nothing.
[...]

Issue #997, opened 2026-09-11
merge-forecast: certifiedWidth is computed from wave 1 only, so --width fails for any foundation-first DAG


voodootikigod:
## Summary

`certifiedWidth` is computed from **wave 1 only**. For any DAG with a foundation ticket — the shape P2 doctrine actively recommends — wave 1 holds a single ticket, so `certifiedWidth` is `1` no matter how wide the graph gets afterwards. `--width` then fails for exactly the decompositions the ADLC tells you to write.

## Mechanism

`packages/merge-forecast/lib/forecast.mjs:134-136`:

```js
// CertifiedWidth: greedy largest set of pairwise-below-threshold tickets in wave 1
const wave1 = waves[0] ?? [];
const certifiedWidth = computeCertifiedWidth(wave1, pairResults, conflictThreshold);
```

`computeCertifiedWidth(wave1Ids, …)` (`:203`) only ever considers pairs where **both** endpoints are in `wave1Ids` (`:210`). Later waves are never scored.

Two consequences:

1. **The gate.** `:154` — `if (width !== null && width > certifiedWidth)` pushes a gate failure. A 1 → 4 → 1 DAG (foundation, four parallel leaves, integration) has a true maximum concurrency of 4 and a `certifiedWidth` of 1, so `--width 4` fails.
2. **The recommendation inherits the ceiling.** `:145-148` — `recommendedWidth = min(certifiedWidth, backpressureWidth, width)`, so every downstream consumer reading `recommendedWidth` also sees 1.

## Why this looks like a defect rather than a deliberate narrowing

`ADLC.md` D2 describes width as a property of **the schedule**, not of its first wave:

> **Tool: `merge-forecast`** — validates the ticket DAG (declared scopes, no orphan edges), scores pairwise conflict probability (signals 1–3), computes float per ticket (feeds `model-router`), emits the dispatch schedule + width recommendation. Exit 2 = partition unsafe at requested width.

and the width heuristic it defines — `width ≈ mean ticket build time ÷ mean merge-rebase-regreen time` — is a steady-state property of the whole run. Wave 1 of a foundation-first DAG is not steady state; it is the serialized prologue that D2 explicitly tells you to build and merge first:

> a contract is pinned when the **foundation is built first and merged to main** … *before* the fan-out

So the tool currently certifies the narrowest moment of the schedule and reports it as the schedule's width.

## Impact

[...]

Issue #996, opened 2026-09-11
docs: handoff gate documented as Enforcing on Bash|Shell after #966, and the deny recovery path (handoff doctor --clear) is undocumented


voodootikigod:
## Summary

Two documentation gaps around the context-rot handoff gate, both hit in the field today:

1. `docs/integrations/claude-code.md` documents the handoff gate as **Enforcing on `Bash|Shell`**, which has been untrue since 1.11.1 disconnected it (#966).
2. **There is no documented recovery path** for an operator whose session is wedged by a handoff deny — and the wedge denies *every* Bash call, including read-only ones.

## 1. Stale integration table

`docs/integrations/claude-code.md:144`:

> | **context-handoff** | PreToolUse | **Enforcing** (T157 / context-rot handoff slice 4): … Denies Edit/Write/MultiEdit/NotebookEdit **and Bash\|Shell** under an active deny-set (fail-closed-all); …

As of 1.11.1, `plugins/adlc-claude-code/hooks/hooks.json` wires no `handoff` verb at all and has no `PreToolUse` matcher on Bash. The row describes 1.11.0 behavior. #966's changelog entry covers claude-code, codex, pi and opencode, so the other integration docs should be checked for the same row.

## 2. No documented recovery path

Searching `docs/` and `README.md` for `ADLC_HANDOFF`, `handoff bypass`, and `handoff doctor` returns **zero hits**. The operator-facing surface for recovering from a deny is entirely undocumented.

This matters more than a normal doc gap because of how the failure presents. With an active deny, every Bash call fails closed:

```
mutation denied (D3:unauthorized_open:<session-id>, bash_fail_closed_under_deny).
Resume via host `adlc handoff resume` / repair, or continue in a fresh session.
Agent Shell cannot clear deny-set.
```

`git status` is denied. `ls` is denied. The agent cannot diagnose its own wedge, and the message's two suggestions (`handoff resume`, "continue in a fresh session") do not resolve the **orphaned-unbound** case — a deny marker with `ticket_id: null` and `content_hash: null`, which nothing can ever consume.

The actual remedy exists and is good — it is just undiscoverable:

```sh
adlc handoff doctor                          # lists orphaned-unbound markers + their writers
ADLC_MANIFEST_KEY=… adlc handoff doctor --clear --write
```

I found it only because `doctor`'s own stderr names it. It should be in the docs.

[...]

Issue #995, opened 2026-09-11
context-handoff: source comments and ENFORCING_MODES still claim the handoff gate is wired after #966 disconnected it


voodootikigod:
## Summary

PR #966 (shipped in 1.11.1) disconnected the context-rot handoff gate by removing its entries from each harness plugin's `hooks.json`. The enforcement **code** was left fully intact — which is fine — but the code still *describes itself* as wired and enforcing. Anyone reasoning from the source (human or agent) will conclude the handoff gate is live. It is not.

## Evidence (verified at `main`, commit `c410918c`)

`plugins/adlc-claude-code/hooks/hooks.json` wires exactly six verbs and **no** `handoff`:

- `PreToolUse` → `Edit|Write|MultiEdit|NotebookEdit`: `rails`, `buildgate`
- `SessionStart`: `preflight`, `context`
- `PreCompact` / `PostCompact` / `SubagentStart` / `SubagentStop`: `context`
- `PostToolUse` → `Edit|Write|MultiEdit|NotebookEdit|Bash`: `flail`
- `Stop`: `manifest`, `review`

The code asserting otherwise:

| Location | Claim | Status |
|---|---|---|
| `plugins/adlc-claude-code/hooks/adlc-hook.mjs:11` | `handoff (PreToolUse) → context-rot deny/handoff (D1–D3) [ENFORCING]` | stale — nothing dispatches this mode |
| `plugins/adlc-claude-code/hooks/adlc-hook.mjs:2624` | "The `rails`/`buildgate`/`handoff` modes are ENFORCING — a crash must FAIL CLOSED" | stale for `handoff` |
| `plugins/adlc-claude-code/hooks/adlc-hook-run.mjs:23` | `ENFORCING_MODES = new Set(['rails', 'buildgate', 'handoff'])` | reserves a deny-on-timeout / deny-on-crash budget for a mode that can never fire |
| `plugins/adlc-claude-code/hooks/adlc-hook-run.mjs:84` | `handoff: 10_000, // hooks.json timeout minus buffer — enforcing: deny on timeout` | references a `hooks.json` timeout that **no longer exists** |
| `packages/context-handoff/lib/adapter.mjs:673-674` | "only Edit/Write/MultiEdit/NotebookEdit/Bash are wired to it — see hooks.json" | factually wrong as of 1.11.1 |
[...]

PR #1004, opened 2026-09-13
docs: evaluate the antigravity-booster pack for ADLC adoption, and park the decomposition compiler spec

voodootikigod:
Docs only. Two artifacts, no code, no ticket, no manifest evidence.

## 1. `docs/intent/booster-adoption.md` — the booster evaluation

An evaluation of `antigravity-booster` for adoption into the ADLC, answering for each candidate: does it apply to all coding agents, does it belong in a harness plugin, does it violate our doctrine, and what does it buy.

**The headline is a reframe.** Booster is a **client** of this toolkit, not a fork — it shells out to nine `adlc` tools (`model-router`, `merge-forecast`, `hollow-test`, `flail-detector`, `consensus-fix`, `gate-manifest`, `rails-guard`, `review-calibration`, `--version`). So the interesting question is not what booster has; it is what booster had to build *around* our tools, because that negative space is the gap.

Four suspected gaps were checked and withdrawn — where we overlap, ours are generally stronger:

| Suspected gap | Reality |
|---|---|
| No edge-parallax | Implemented — `packages/parallax/bin/parallax.mjs:158-210` |
| Fleet doesn't sandbox gates | Implemented and ahead — `bwrap --unshare-net` plus an egress plane; booster only denies network |
| Ticket-id case collisions unguarded | Guarded at both store backends |
| No worktree/merge/revert lifecycle | Implemented and ahead — `revertMerge` quarantines the branch rather than assuming a reset worked |

What survives as genuinely worth taking, with placement and doctrine notes, is in the document. The most portable item is not a feature but a **gate failure-policy table**: booster makes a differentiated, deliberate decision per tool about what happens when a gate cannot run — `rails-guard` fails **closed**, `flail-detector` fails **open**, the rest degrade to advisory. We have the producer-side exit-code convention in `CONVENTIONS.md` and nothing written for the consumer side. The extractable principle: *a gate that protects an invariant fails closed; a gate that optimizes a heuristic fails open; a gate that records evidence degrades.* The Goodhart Annex covers agents gaming gates; nothing covers gates being absent.

Two findings from the comparison are filed separately as #995/#996 and #997.

## 2. `.adlc/specs/decompose.md` — parked, and the record of why

The one real capability gap was a P2 decomposition compiler: every DAG *validator* ships, nothing *authors* the DAG. This spec went through four adversarial premortem rounds and is **parked, not approved**.

Each round the defects moved rather than reduced — v2's in the gate-orchestration loops, v3's in the store write's trust-root contract, v4's in the validation rules themselves. The recurring killer was identical every time: **the emitted DAG is not applicable.**

That is the diagnosis. This specifies a producer for a consumer that does not exist — there is no atomic multi-ticket write path (#1003), and `adlc ticket create --input` takes one ticket while persisting a supplied id verbatim (`service.mjs:106`), so a provisional `T1` collides with the 96 legacy `T<n>` shards already in the store. With no ground truth to validate against, every round re-litigated what the DAG should look like.

It unblocks when #1003 lands. §11 records the round-four findings verbatim so they are not rediscovered — including three that are defects in shipped code rather than in the spec:

- **`merge-forecast` filters candidate pairs to the same wave** (`forecast.mjs:170-175`) before vetoing, while `parallelEligiblePairs` does not — so any caller assuming those agree is stricter than the gate.
- **`validateTicket` requires only `id` and `title`** (`schema.mjs:4-12`), not `body` or `scope`. Combined with `scopesOverlap` reading `a.scope ?? []`, a scope-less ticket passes structural validation and is then hard-failed by coldstart.
- **`fence()` is forgeable always, not only on truncation.** The tag is `${label}-${capped.length}` (`core/lib/text.mjs:48`) — a pure function of content the author controls. `tail()` also keeps the *last* `maxChars`, so an over-cap spec silently loses its opening sections.

## Store state

Unchanged. No ticket was created and no manifest evidence recorded — the anchor ticket authored against v2 declared rails over four files for work no longer planned, and its `spec-lint` evidence was bound to a spec hash two revisions stale, so it was already unusable. Both artifacts were untracked and removed. `adlc ticket store status` reports the same 275 tickets and the same `storeHash` `9c1ece3a…` as before this branch.

## Related

- #995, #996 — handoff gate drift after #966 (source and docs)
- #997 — `merge-forecast` certifies width from wave 1 only
- #1003 — no atomic multi-ticket write path (parks the spec above)


voodootikigod:
The `fence()` finding described in this PR is now reproduced and filed as **#1005** (P0, security): the tag is `${label}-${capped.length}`, a pure function of content the author controls, so fenced content can forge its own closing marker. Not conditional on truncation. `694daec7` updates the spec to cite it.

Reproduced against the real `ISSUE_BODY_FENCE_CAP = 8000` — the sharpest exposure is `autopilot/lib/triage.mjs:83`, which fences a GitHub issue title and body, i.e. text any outside contributor controls.

PR #1014, opened 2026-09-14
fix(cursor): make plugin MCP launcher self-contained 🧡

rebelliard:
## 🤓 Summary

The Cursor plugin MCP could not start or bind reliably in a real Cursor install
for our internal users at Mentimeter.

```text
Error: Cannot find module
'/path/to/repos/agents/bin/adlc-mcp-wrapper.mjs'

code: 'MODULE_NOT_FOUND'
MCP error -32000: Connection closed
```

The path changed between reloads — first `~/bin/...`, then the workspace
`bin/...`. Cursor was resolving the plugin’s relative launcher against its
current `cwd`, not the plugin directory.

This PR anchors and bundles the launcher, then hardens the Roots binding flow.
It does **not** claim MCP is fully shipped yet.

## 🛠️ Design

- Launch the bundled proxy from `${CURSOR_PLUGIN_ROOT}`, with the plugin root as
  its `cwd`. The plugin no longer needs its own `node_modules` or a workspace
  shim.
- Bind `adlc mcp-server` only after Cursor initializes and supplies Roots. The
  proxy accepts Cursor’s bare absolute-path Roots and `file://` URIs.
- Fail pending work explicitly when Roots fail or change. Ignore stale Roots
  replies, do not mix ambient workspace environment variables into real Roots,
  and retire the old child before rebinding.
- Generate the bundle through one build path used by local builds, freshness
  checks, and releases.
- Keep `packages/cli/lib/mcp-server.mjs` unchanged. The Cursor wrapper selects
  the correct consumer `cwd`; it does not alter the shared MCP server.

### 🔬 Installed-Cursor proof

Tested on Cursor Desktop 3.20.10 with no plugin-cache `node_modules` and no
workspace shim:

- `${CURSOR_PLUGIN_ROOT}` expanded in both `args` and `cwd`; the bundle booted
  from the plugin root.
- Cursor returned a bare absolute path from `roots/list`; the proxy bound the
  workspace, exposed `adlc_gate` and `adlc_prosecute`, and
  `adlc_gate gate-manifest show` returned `ok` with `exitCode: 0`.

Regression coverage includes initialize ordering, queued tool requests, Roots
errors and stale replies, root changes, POSIX and Windows Root paths, child
shutdown, bundle freshness, and release-time regeneration.

Focused MCP, packaging, release, and install-smoke checks pass. Preflight was
not re-run for this update.

## ⚠️ Still unproven live

- `roots/list_changed` rebind: Cursor 3.20.10 advertises `listChanged: false`.
- Multi-root refusal: covered by tests, but not yet exercised in Cursor.


voodootikigod:
Reviewing right now.

voodootikigod:
## Maintainer review — accepted, with three small follow-ups pushed to this branch

Thanks for this, Rafael. The installed-Cursor evidence in the description matches what I could reproduce locally, and it's refreshing to get a PR that scopes its own claims this carefully. Overall verdict: **ship it**. I found three issues, all small, and pushed fixes directly (maintainer edits were enabled). Details below so you can see exactly what changed and why.

### What I verified

| Check | Result |
|---|---|
| Full `npm test` at `48ddee44` (your head) | ✅ 56/56 segments green; 57/57 after the follow-ups at `a353aedb` |
| `plugins/adlc-cursor` tests | ✅ 227/227 |
| `release`, `ticket-store-boundary`, docs `integration-facts` tests | ✅ 53/53 |
| `scripts/cursor-install-smoke.mjs` | ✅ PASS |
| Committed bundle + metadata vs `buildCursorMcp({ write: false })` | ✅ byte-identical, both at your head and after merging current `main` |
| **Standalone boot**: only `bin/adlc-mcp-wrapper.bundle.mjs` copied into an empty dir (no `node_modules`, no `package.json`), cwd = that dir | ✅ `initialize` → `notifications/initialized` → `roots/list` (bare absolute path) → queued `tools/list` answered with `adlc_gate` + `adlc_prosecute`. With a non-ADLC root it fails closed (`UNRESOLVED`) and the queued call gets a JSON-RPC error |
| Planted defect: `binding = false` at spawn (forward client calls before the child handshake completes) | ✅ killed by 2 behavioural tests, plus the bundle-freshness tests |
| `rails-guard-ci` on a simulated merge ref | ✅ only the expected trust-root authorization for `package.json` (see "Remaining steps") |
| Cross-model adversarial review (Antigravity / `gemini-3.7-flash-medium`) | ✅ approve (details below) |

**How the cross-model review was run.** The 334 KB diff is bigger than the agy CLI's prompt limit (~130 KB), so I split it into a chain of scoped commits: source/build/trust-root → tests → docs/spec → generated bundle. The last commit's tree is byte-identical to this PR's tree, so the chunks cover the whole PR with nothing missing. The generated bundle was not sent to the LLM because the freshness test proves it byte-equal to the builder output.

- Source, build, release and trust-root files: **approve, 0 findings**
- Docs, spec, ADR, skills: **approve, 0 findings**
- Tests: **approve**, with 5 advisory findings. Each is adjudicated below with evidence.
- My follow-up commit (below): **approve, 0 findings**

### Issues found and fixed (commit a353aedb, after merging `main` in a0dacad1)

1. **Test temp-dir leak.** Both `mcp-wrapper.test.mjs` (this pattern predates the PR) and the new `mcp-roots-proxy-lifecycle.test.mjs` created a fresh `adlc-mcp-state-*` directory under `tmpdir()` on every wrapper launch and never removed it. Measured: **14 leaked directories per run** of those two files. The machine I reviewed on had 473 of them. This repo has already had one `/tmp` inode-exhaustion incident from leaked fixtures. **Fix:** each launch registers its dir, and an `after()` hook removes them. Re-measured: 0 leaked.
2. **Misleading failure when the bundle goes stale.** The bundle inlines `@adlc/tickets` and `@adlc/core`, so any future edit to a bundled module makes it stale. `ticket-store-boundary.test.mjs` then reports the bundle as a *direct ticket-store writer*, which reads like a security regression in a file the author never touched. `packaging.test.mjs` does say "run npm run build:cursor-mcp", but that is a different test and easy to miss. **Fix:** the boundary failure now explains the stale bundle and names `npm run build:cursor-mcp`. A new test asserts the hint appears only when the bundle is listed, so an ordinary direct writer is never misdiagnosed as a stale bundle. Verified by appending one byte to the bundle: the new message appears.
3. **Branch was behind `main`.** The `main` ruleset requires branches to be up to date. **Fix:** merged `main` in (no conflicts). The bundle is still byte-fresh after the merge, since nothing on `main` touched the bundled modules.

### Cross-model findings adjudicated with no code change

- **"Hollow assertion in the child-handshake collision test"** (high). **Refuted.** With the handshake-queue mutant planted, that exact assertion fails (`client calls must queue until the child handshake replies`, `3 !== 1`). The in-process harness dispatches synchronously, so the assertion is load-bearing.
- **"Grandchild fixture server orphaned when a test fails"** (high). **Refuted.** On your original code I planted a `throw` right after the server pid is read in the SIGKILL-fallback test. **0** fixture servers survived. The fake server exits when its stdin closes, and its SIGTERM listener does not keep the event loop alive. I prototyped a pid reaper, then deleted it because it guarded nothing.
- **"SIGTERM handlers / `process.kill(pid, 0)` / `symlinkSync` fail on Windows"** (medium). **Not applicable today.** The plugin tests run only on the ubuntu CI matrix; the Windows leg of `ticket-store-platform` runs only the `@adlc/tickets` workspace. This becomes real if these tests ever join a Windows matrix.
- **"75 ms negative wait in the `list_changed` test is flaky"** (low). **Not a flake.** Under load it can only produce a false *pass*, never a spurious red, and the assertions after it independently check that id 22 is answered exactly once. Left as is.

### Non-blocking notes (no change requested)

- **Ongoing cost of committing the bundle.** Cursor installs straight from git, so committing the bundle seems forced. But `workspace-resolve` pulls a large part of `@adlc/tickets` (manifest segments, trust-root, evidence, lock) and `@adlc/core` (llm, git, risk-tier) into it. Every future PR touching those modules must now carry a ~120 KB regenerated diff, and that diff alone nearly fills the agy review budget. It may be worth a follow-up to shrink the bundled surface to what the proxy actually needs.
- **CLI resolution order changed on POSIX.** The global install next to the running `node` (and npm prefix / `NODE_PATH` / Volta / pnpm) now wins over `adlc` on `PATH`. Someone with two installs could have MCP running a different CLI version than their shell. The jsdoc documents it; a README troubleshooting line might save a support question.
- **No retry after a roots timeout.** A single `roots/list` timeout (10 s) leaves the proxy unbound for the rest of the session. It fails closed, so this is safe, and it's reasonable while Cursor advertises `listChanged: false`.

### Remaining maintainer steps (not code)

- ✅ Pending fork-PR workflow runs approved.
- ✅ `trust-root-change` label applied. `rails-guard` still needs a **non-author CODEOWNER approval**, because `package.json` and `package-lock.json` are trust-root files.
- ✅ **Cross-model attestation recorded** (commit 4f74703c): signed `cross-model-review` approve under ticket T65, provider `gemini`, author `anthropic`, for revision `git-change:3e0c988b6b6d316175ad9d9254683e5a1c4e545a:afb6df1efc902e414657c6a80a68025f6ce5b507029f889c29261085642607ac`. A local `tier-check --author-provider anthropic` PASSES against it. The full suite on the final head is also green (57/57 segments). Any further rebase or content change moves the revision and requires a fresh attestation.

voodootikigod:
## Re-review of the follow-up commits (`c7f9c4f0..56351012`) — two fixes pushed, re-attested

@rebelliard — your five commits this morning changed the reviewed content, so the earlier cross-model attestation (bound to revision `…:afb6df1e…`) no longer matched and `gate` correctly went red. That's the gate working as designed, not a problem with your change. I re-reviewed the delta, pushed two fixes (one source, one test flake), merged `main`, and re-attested the new head.

### What I checked on your delta

| Check | Result |
|---|---|
| Affected suites (`plugins/adlc-cursor`, `build-cursor-mcp`, `cursor-install-smoke`, `mutation-gate`, `ticket-store-boundary`, `release`) at `56351012` | ✅ 342/342 |
| Cross-model review of `4f74703c..56351012` (Antigravity / `gemini-3.7-flash-medium`, 47 KB, fully inlined) | gate-approve, but the model's verdict was **needs-attention** with one finding — adjudicated below as real |

The changes read well. Waiting for the real `exit` before binding a replacement, `childAcceptingRequests` so a closed stdin is never written to while buffered replies still drain, and handling child-stdin `error` all close genuine races, and the new tests pin each one.

### Issue found and fixed (commit 643aa1ec)

**Child retirement could hang forever after SIGKILL.** Previously the SIGKILL timer called `finish()` unconditionally, so retirement was bounded. The new code resolves only when the child emits `exit`/`error` after a successful `kill("SIGKILL")`. That is right for the normal case, but a child in uninterruptible I/O (a stuck NFS or FUSE read, for example) accepts SIGKILL and emits no `exit` until the kernel releases it. Meanwhile:

- `rebindTail` never proceeds, so `roots/list` is never re-requested and any `tools/call` queued during the rebind **never gets a reply**;
- `shutdown()` never resolves, so the wrapper lingers after Cursor closes stdio.

**Fix:** after SIGKILL, wait for the real exit but at most `killTimeoutMs` (default 2 s, a new fifth parameter to `retireChildProcess`). If the deadline passes, log `child <pid> did not exit …ms after SIGKILL; continuing` and resolve. Continuing is safe: a SIGKILLed process never returns to user space, so it cannot write MCP traffic or touch `.adlc/` alongside its replacement. Both retirement timers (SIGTERM→SIGKILL and the new kill deadline) are now tracked, so neither leaks into `retirementTimers`.

- New test `SIGKILL retirement is bounded when the child never reports exit`: **RED** on your head ("retirement never resolved"), **GREEN** with the fix. It also asserts both timers are released and the `exit`/`error` listeners are removed.
- Your test `SIGKILL retirement waits for the child exit event before resolving` still passes unchanged; the default 2 s deadline never fires in the normal path.
- Bundle regenerated with `npm run build:cursor-mcp`; freshness tests pass.
- Affected suites 343/343; full suite on the final head `08e80980`: 57/57 segments green.
- Cross-model review of this fix: **approve, 0 findings** (the model verdict agrees; no suppressed findings).

### Second issue found and fixed: a load-dependent flake (commit 08e80980)

After merging `main`, `a live silent child times out queued calls without overlap or orphaning` failed in a parallel run of the cursor and scripts suites (1,315 tests). The proxy retires the silent fake child 50 ms + 20 ms after spawn, but the test waited for a **pid file the child writes after Node starts up**. Under load, Node startup outlasts 70 ms, so the child is killed before it writes the file and `waitFor(existsSync(pidFile))` times out.

- Reproduced with every core busy: **0/8** passing on the merged head, and **5/6 failing on `56351012`**, so it predates my follow-up commit.
- **Fix (test-only):** take the pid from the spawned `ChildProcess` in `spawnImpl`. Every assertion is unchanged (timeout reply `-32001`, exactly one spawn, the process is gone after retirement). SIGKILL escalation keeps its deterministic coverage in the tests that wait for the pid file *before* signalling.
- Under the same load: **8/8** passing.

This matters because `test (18/20/22)` are the required checks, and CI runners are slower than a dev box.

Also merged `main` again (`561cc594`, #1016 backlog-groom; no overlap with this PR, bundle still byte-fresh), since the ruleset requires the branch to be up to date.

### Attestation

✅ Re-recorded (commit ebb6745b): signed `cross-model-review` approve under T65, provider `gemini`, author `anthropic`, for revision `git-change:6280f8b5bc627dd56676e1e3847cb810722528c0:493565eb15e9e0924af3ebd616e81010bc960fb0f54317d270ba78bc5ef56ba3`. It is the second entry in the branch segment's hash chain, and a local `tier-check --author-provider anthropic` PASSES. Every delta since the first attestation was cross-model reviewed: your five commits, my SIGKILL fix, and the flake fix.

**About the red `test (18)` on `56351012`:** that is this same flake. The Cursor install smoke runs the plugin unit tests under load, and `a live silent child times out queued calls…` timed out waiting for the pid file. It is fixed at the new head.

**Heads-up:** any further push changes the revision again and needs a fresh attestation. If you have more changes planned, it's cheaper to land them first and ping me, and I'll re-attest once at the end.

rebelliard:
@voodootikigod I pushed some commits with the intention of fixing some CI issues. I'll stop doing so unless told to do so. 😄

voodootikigod:
Either way works for me @rebelliard happy to wait until you give me a go

rebelliard:
@voodootikigod Go 😃

voodootikigod:
## Round 3: `a93cb254` reviewed and re-attested; the `mutation-gate` red was base drift

@rebelliard — the atomic-pid commit is a good catch and reviewed clean. Pushed `main` + a fresh attestation on top (`4a7ac3c7`).

**Your commit `a93cb254`** — cross-model review (Antigravity / `gemini-3.7-flash-medium`): **approve, 0 findings**. Publishing the pid by same-directory rename makes the file all-or-nothing, and `readChildPid` closes the hazard that mattered: a torn read gives `Number("") === 0`, and `process.kill(0, 0)` probes *this process group* instead of reporting `ESRCH`, so a child that outlived the wrapper would have read as dead. Assertions unchanged.

**The red `mutation-gate` was not a defect in this PR.** Its survivor was `packages/lesson-foundry/lib/llm.mjs:22` — a file this branch never touches. The branch had fallen behind `main`, so the gate's changed-file set picked up unrelated commits from `main`. After merging `main` (`808dfe46`) the set is this PR's files only, and locally the gate is **exit 0 with all 16 drawn mutants KILLED**:

```
mutation-gate: 12 file(s) covered by fast test targets
KILLED  plugins/adlc-cursor/lib/mcp-roots-proxy.mjs …
KILLED  plugins/adlc-cursor/lib/mcp-proxy-runtime.mjs …
KILLED  scripts/build-cursor-mcp.mjs …            (16/16)
```

**Also verified on the merged head (`c851fb02`):** full `npm test` **57/57 segments green**; bundle and metadata still byte-identical to the builder (worth checking, since `main` brought a `packages/core` change and the bundle inlines part of that package — this time nothing reachable from the proxy changed, so no regeneration was needed).

**Attestation** ✅ re-recorded (`4a7ac3c7`): signed `cross-model-review` approve under T65, provider `gemini`, author `anthropic`, revision `git-change:808dfe46d6af06d128fda0963c084b62f9e797ff:8dc4e41440680d4b5d472c45f37ee2ecee45e62c9f736ec25843e514c3ec20ac`. Third entry in the branch segment's hash chain; local `tier-check --author-provider anthropic` PASSES.

### Where the PR stands

- ✅ `test (18/20/22)` (the required checks), `ticket-store-platform` (9 jobs), `actionlint`, `router-drift`, `pre-ga-gate`, live-latest
- ✅ `gate` — once CI re-runs on this head
- ✅ `mutation-gate` — should clear now that the branch is current with `main`
- ⏳ `rails-guard` — needs a non-author CODEOWNER approval (trust-root files `package.json` / `package-lock.json`); the `trust-root-change` label is applied. This one is a human gate, not something a push can fix.

From my side the change is ready. Each push re-opens the attestation, so if you have anything else planned, land it and ping me and I'll re-attest once at the end.

rebelliard:
@voodootikigod Nothing else on my end! 🙏

voodootikigod:
## Round 4: the `mutation-gate` red was real — coverage gap closed

Unlike the previous one (base drift), this survivor was a genuine gap in a file this PR owns.

**`scripts/build-cursor-mcp.mjs:8` `[value-substitute]` SURVIVED** — the `ROOT` constant. Every test passes an explicit `root`, so the module's own default, the one `npm run build:cursor-mcp` and `releaseMain` rely on, was never exercised: mutating it changed nothing any test noticed. My earlier local run drew a different mutant on that line and killed it, which is why this only surfaced in CI.

**Fix (`02702568`, test-only):** assert that a no-argument dry run produces the same paths and bytes as an explicit repo-root build. The expected root is computed independently in the test, so it cannot pass vacuously. Verified it bites: with `ROOT` resolved to `"."` instead of `".."`, the new test fails.

**Verified on the merged head:**

- local `mutation-gate` → **exit 0, 20/20 drawn mutants KILLED**, including `scripts/build-cursor-mcp.mjs:8 [value-substitute]`
- full `npm test` → **57/57 segments green**
- cross-model review of the new test (Antigravity / `gemini-3.7-flash-medium`) → **approve, 0 findings**
- merged current `main` (`7b261c1e`); attestation re-recorded for revision `git-change:7b261c1efe5e2212bdae9915806748831510dcca:39777a82aab28a20350554fd179db0e957a94cb37c92062df56b3956a931aabb`, `tier-check` PASSES

### Remaining

Only `rails-guard`, which needs a non-author CODEOWNER approval for the trust-root files (`package.json`, `package-lock.json`). The label is applied; this is a human gate. Note it does not re-run on review submission (`ci.yml` triggers on `push` / `pull_request` only), so the job gets re-run once the approval is in.

One process note worth recording: merging `main` re-opens the attestation even when nothing in the PR changes. The change-set digest is base-independent — it stayed `8dc4e414…` across one such merge — but the gate matches the full `git-change:<base>:<digest>` string, so a new base means a new signature. With `main` moving several times a day, the last mile wants an approve-then-merge window rather than another round trip.

voodootikigod (approved):
Reviewed 🥇

Issue #1013, opened 2026-09-14
mutation-gate: a diff with no mutable lines can never pass, and there is no way to supply the mutant by hand


voodootikigod:
## Summary

`scripts/mutation-gate.mjs` fails closed when a diff's changed lines produce no mutants. That behaviour is correct and deliberate (#658 — a zero-mutant file means the changed code was never verified, so reporting a pass would be a false green). But it leaves a class of legitimate changes with **no path to a green gate at all**, and no sanctioned way to record why.

## How it shows up

PR #1011 changes exactly two source lines in `packages/core/lib/text.mjs`:

```diff
+import { randomUUID } from 'node:crypto';
-  const tag = `${label}-${capped.length}`;
+  const tag = randomUUID();
```

```
error: diff-derived file(s) produced zero mutants — packages/core/lib/text.mjs: no mutable
line was found (comment-only, blank, or a shape none of the mutation operators recognize).
The changed code was never actually verified; refusing to report a pass.
```

Neither line is mutable by the current operators: an `import` statement, and a bare assignment with no conditional, no binary operator, and no returned expression to null out. The gate is right that *it* did not verify the change. It is wrong to conclude nothing did.

## Why this is a gap rather than just strictness

The change is verifiable, and was verified — just not by this tool. The one mutant that matters is "revert the tag to the length-derived form", which four test files kill:

| Test file | Exit with mutant |
|---|---|
| `packages/core/test/text.test.mjs` | 1 |
| `packages/fleet/test/adapter-shared.test.mjs` | 1 |
| `packages/autopilot/test/triage.test.mjs` | 1 |
| `packages/review-calibration/test/judge-bounding.test.mjs` | 1 |

That is exactly what the gate exists to establish. It could not generate that mutant because "replace an expression with a different expression of the same shape" is not one of its operators.

The affected class is not exotic: swapping one call for another, changing a constant's provenance, replacing a template literal with a function call. Security fixes are over-represented in it, because they often *are* a single substitution.

## Options

1. **A semantic mutation operator** — "replace the right-hand side of an assignment with a neighbouring plausible value". Broadest fix, most work, and the LLM-authored-mutant path in `hollow-test` already gestures at it.
2. **A declared-mutant escape hatch** — let a ticket or the PR declare a hand-authored mutant (a patch plus the test command expected to fail), and have the gate apply it and assert it is killed. Keeps the burden of proof on the change, does not weaken the gate, and mechanises what #1011 currently does by hand in prose.
[...]

PR #1015, opened 2026-09-14
fix(prompts): fence the four builders that embedded externally-authored content raw (#1010)

### LINKED ISSUE #1010: prompts: four builders embed externally-authored content without fence() (spec-lint, lesson-foundry, premortem, consensus-fix)

voodootikigod:
Closes #1010. **Stacked on #1011** — base is `fix/1005-fence-nonce`, which supplies the completeness sweep this builds on. Merge #1011 first.

## What was wrong

Four prompt builders spliced text they did not author straight into an LLM prompt:

| Builder | Embedded content | Who authors it |
|---|---|---|
| `spec-lint/lib/llm.mjs` | acceptance-criterion text | whoever filed the work |
| `lesson-foundry/lib/llm.mjs` | prosecution findings + cluster name | model-authored text fed back into a model |
| `premortem/lib/prompt.mjs` | **the entire spec file**, unfenced and placed *before* the instructions | whoever filed the work |
| `consensus-fix/lib/prompt.mjs` | raw failing-test output and each source excerpt | the code under test, behind ``` only |

The `consensus-fix` case had the weakest protection rather than none: a triple-backtick block is escaped by writing a triple backtick, which test output can trivially contain.

These were found by the completeness sweep in #1005, which forced a classification of every prompt-sending module in the suite. They are the reason that sweep shipped with an `UNGUARDED_REVIEWED` list.

## Two choices worth reviewing

**`lesson-foundry` fences the cluster name and the samples JSON separately.** One fence around both would let a hostile name break out through the samples block, or vice versa.

**`premortem` uses a 64000 cap where the others use 8000.** The spec *is* the payload there, not one field among many, so any cap truncates the artifact under review — and `fence()` is tail-biased, so an over-cap spec would silently lose its **opening** sections. That is #1007. The constant carries the reasoning as a comment rather than being a bare number. I deliberately did **not** reorder the prompt to put the spec last as a workaround; fencing already keeps the instructions outside.

## The sweep

All four move from `UNGUARDED_REVIEWED` into `GUARDED`, each with `mustContain` naming its `fence()` call and `mustNotMatch` naming the raw interpolation it replaces. The allowlist now has **zero** "Recorded, not accepted" entries.

Re-verified the sweep still goes red on a planted prompt-sender in a sibling package, and green once removed.

## A stub that could have hidden the bug

`packages/premortem/test/premortem.test.mjs` runs `run.mjs` in a subprocess under a loader that stubs `@adlc/core`. The stub is hand-written and lacked `fence`, so the change broke it. It now re-exports the **real** `fence` from `core/lib/text.mjs` rather than a hand-written one — a fake would let the stub make an unfenced prompt look fenced. Only the LLM surface stays mocked.

## Evidence

Red tests written first — one file per builder, feeding a hostile payload containing a forged terminator and asserting it stays inside the fence. All four failed before the change:

```
spec-lint        exit=1  RED
lesson-foundry   exit=1  RED
premortem        exit=1  RED
consensus-fix    exit=1  RED
```

Reverting each `fence()` call by hand kills its test:

| Package | Mutant exit | Verdict |
|---|---|---|
| `spec-lint` | 1 | KILLED |
| `lesson-foundry` | 1 | KILLED |
| `premortem` | 1 | KILLED |
| `consensus-fix` | 1 | KILLED |

| Gate | Result |
|---|---|
| `npm test` | **56/56 segments**, exit 0 |
| `adlc rails-guard --base fix/1005-fence-nonce --ticket …` | **exit 0** |

Note on rails-guard: run against `origin/main` it reports `[rail-edit] packages/core/lib/text.mjs`, which is a **stacking artifact** — that file is changed by the parent branch (#1011), not by this one. `git diff fix/1005-fence-nonce...HEAD` does not contain it. Against the correct base it is clean, which is why the PR base is the parent branch.

## Out of scope

`packages/core/lib/text.mjs` is a declared rail on this ticket — this work consumes the fence primitive, it does not change it. `tail()`'s truncation direction is #1007; the mutation-gate zero-mutant gap is #1013.

Ticket `T-01M2GZNJFV3RZV0T7ZKGJ2BE62`, coldstart-gated with zero gaps.


voodootikigod:
## Rebuilt on main after #1011's squash-merge

#1011 was squash-merged, so this branch still carried its pre-squash commits and the PR went `DIRTY`. Rebuilt per the squash-merge recovery: reset to `origin/main` and cherry-picked the single unique commit rather than rebasing (a rebase would have replayed #1011's commits). The diff against main is now exactly the #1010 work.

## The mutation gate found real gaps in this PR

With #1020's `value-substitute` operator on main, the gate now runs here and it caught six survivors — all mine, all `off-by-one` on the cap constants I introduced:

```
SURVIVED  packages/consensus-fix/lib/prompt.mjs:12   const TEST_OUTPUT_MAX_CHARS = 4000;
SURVIVED  packages/consensus-fix/lib/prompt.mjs:13   const FILE_EXCERPT_MAX_CHARS = 8000;
SURVIVED  packages/lesson-foundry/lib/llm.mjs:9      const CLUSTER_NAME_MAX_CHARS = 200;
SURVIVED  packages/lesson-foundry/lib/llm.mjs:10     const SAMPLES_MAX_CHARS = 8000;
SURVIVED  packages/lesson-foundry/lib/llm.mjs:29     JSON.stringify(samples, null, 2)
SURVIVED  packages/spec-lint/lib/llm.mjs:7           const CRITERIA_MAX_CHARS = 8000;
```

I checked whether these were equivalent mutants: `isTuningIdentifier` returns **false** for every `*_MAX_CHARS` name, so they are not masked — and a char cap is genuinely observable. The repo's own convention is to assert it (`coldstart.test.mjs` pins `embedded.length === 8000`). So these were real hollow coverage I introduced by adding caps without testing them.

Six assertions added — each cap pinned to its exact value, plus the JSON indent. Final gate run: **8 mutants, 8 killed, 0 survived.**

`npm test`: 57/57. rails-guard-ci: exit 0.

